PR #17, opened 2014-11-21
Add replicate

MichaelXavier:
I started writing an FFI version and realized it was identical to map except for the use of `const`, which didn't see worth optimizing away.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,8 @@
 
     range :: Number -> Number -> [Number]
 
+    replicate :: forall a. Number -> a -> [a]
+
     reverse :: forall a. [a] -> [a]
 
     singleton :: forall a. a -> [a]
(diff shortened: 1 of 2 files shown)


jutaro:
Yes, please add replicate, I need it, You need it, Everyone needs it ....

jacereda:
Yes, please... I would prefer the FFI approach though, creating a single array is better than creating 2 (result+range). I'll send another PR.

paf31:
Can we not use the FFI please? We should be able to construct an efficient `replicate` from what we have already. Either this (just needs rebasing) or `STArray` should work.

jacereda:
This one will allocate twice. I'm not very keen on using the FFI if there's a fast enough alternative, but 2x seems like a high price.

PR #19, opened 2014-11-29
Added array creation and manipulation functions

NightRa:
Added the functions: fill, tabulate, replicate, intercalate, intersperse, break, any and all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@
 
     append :: forall a. [a] -> [a] -> [a]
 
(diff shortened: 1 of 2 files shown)


jutaro:
Everyone adds replicate, but when do we get it?

paf31:
Closing this, since it is out of date, but thanks anyway. As for the more efficient versions of `any`, `all`, etc., I'd suggest creating a separate package. It seems that there is some interest in a library of monomorphic versions of several core library functions for performance reasons, but we don't want to keep duplicate versions of those functions here.

PR #24, opened 2015-02-08
Some optimizations

jacereda:


paf31:
Aren't most of these limited to ES5?

paf31:
Then again, I guess we have polyfills to solve that problem.

jacereda:
Concat should be everywhere.  map/filter are ES5. es5-shim provides good definitions for those preallocating the resulting array. One possibility would be to just write preallocating versions, but AFAIK those are implemented as natives in most (if not all) implementations and it would be a pity to lose that. I think only IE8 would be problematic, and requiring es5-shim for IE8 doesn't seem like a big deal.

garyb:
I think most of these have been integrated or made redundant now through various updates for 0.7, so I'll close this. Thanks!

PR #27, opened 2015-02-18
Added fixity declaration to (..) (range operator)

pseudonom:
I'm not certain exactly what the fixity declaration should be, but it seems that (..) should have one.

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -290,6 +290,7 @@ foreign import range
   \  };\
   \}" :: Number -> Number -> [Number]
 
+infix 8 ..
 (..) :: Number -> Number -> [Number]
 (..) = range
 


paf31:
How does this work with things like `>>=` and `<$>`. Those are my only real use cases where precedence seems important:

```
show <$> 1 .. 5
1 .. 5 >>= \n -> [n, n * 2]
```

should parse as

```
show <$> (1 .. 5)
(1 .. 5) >>= \n -> [n, n * 2]
```

I think it _does_, but could you please check

pseudonom:
Sure. I checked and adding the fixity declaration in the patch makes both of those work.

paf31:
I tested `core-tests` and `test-everything` too, everything looks good. Thanks!

PR #30, opened 2015-03-17
A start on Markdown docs

paf31:
@garyb Could you please review. These aren't really complete. They still need:

- Asymptotics for each function (I need to think a little harder about some of them)
- More examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,27 @@
 
 ## Module Data.Array
 
+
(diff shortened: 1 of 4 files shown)


garyb:
Looks great to me so far. Perhaps worth making it clear in `Unsafe` that "unspecified behaviour" includes runtime errors, but apart from that I have nothing else.

PR #32, opened 2015-03-21
Fix pokeSTArray, some optimizations

jacereda:
It seems pokeSTArray allows poking one element past the end.

### DIFF
--- a/src/Data/Array/ST.purs
+++ b/src/Data/Array/ST.purs
@@ -53,8 +53,7 @@ foreign import peekSTArrayImpl """
   function peekSTArrayImpl(just, nothing, arr, i) {
     return function() {
       var index = i >>> 0;
-      var len = arr.length >>> 0;
-      return index < len? just(arr[index]) : nothing;
+      return index < arr.length? just(arr[index]) : nothing;
     };
   }""" :: forall a h e r. Fn4 (a -> r)
                               r
@@ -70,8 +69,7 @@ foreign import pokeSTArrayImpl """
   function pokeSTArrayImpl(arr, i, a) {
     return function() {
       var index = i >>> 0;
-      var len = arr.length >>> 0;
-      var ret = index < len;
+      var ret = index < arr.length;
       if (ret)
         arr[index] = a;
       return ret;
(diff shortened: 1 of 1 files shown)


src/Data/Array/ST.purs
```diff
@@ -146,12 +141,10 @@ thaw = copyImpl
 foreign import toAssocArray """
   function toAssocArray(arr) {
     return function(){
-      var as = [];
-      var i = -1;
-      var n = arr.length;
-      while(++i < n) {
+      var n = arr.length >>> 0;
+      var as = new Array(n);
+      for (var i = 0 >>> 0; i < n; i++)
```
paf31: What is `0 >>> 0` for here?

jacereda: Just to ensure unsigned values. Probably useless in this case (not in the case of peekSTArray, where it saves a comparison). Do you want me to remove this one?

paf31: Why would `length` be signed?

jacereda: I'm just trying to hint the JIT that both values are unsigned:

    (-1) < ([].length)
    true
    (-1>>>0) < ([].length >>> 0)
    false

I guess this is completely useless since JITs are now quite smart.

jacereda: We really need a good benchmarking library, I'm trying to wrap benchmark.js right now.

garyb: asmjs style? I know `x|0` is recognised as "it's an int" type hint, but not sure if there is one for uints.

jacereda: Looks like >>>0 is indeed the hint for unsigned.

Take a look at the first benchmark in http://mbebenita.github.io/LLJS/

Change the loop limit to something like 0x7fffffff in all 3 tests, it seems int is the fastest type.

jacereda: I've seen very disparate results across browsers/nodes. I'll remove the hints.

paf31:
Looks good to me, thanks. Anyone else care to comment?

garyb:
I don't think it's very likely at all, but isn't `i >>> 0` problematic here? What if you really did have an array that had 4294967295 elements and tried to use `-1` as the index. The original `~~` wasn't much better for similar reasons, and I think that may have been my fault.

Perhaps this is an absurd thing to worry about though.

jacereda:
Should I leave that as normal numbers with 2 comparisons? The important thing here is really the preallocation and the wrong check...

garyb:
It's not something I feel strongly about, just thought it was worth discussing. I'd be happy to merge as it is really. @paf31?

paf31:
:+1:

garyb:
I might try and make an array that large to see how long it even takes to iterate over it...

paf31:
:)

PR #33, opened 2015-03-29
Faster copyImpl using slice().

jacereda:
This one is simpler than the for loop and seems to perform consistently faster across implementations (only Safari seems to remain roughly the same).


garyb:
:+1: :+1: :+1:

PR #37, opened 2015-04-06
Updates

### LINKED ISSUE #36: Should findIndex return Maybe Number instead of failing with -1?
Let me start by saying that I really like how `head`, `tail` and other functions in this module are safe by default (compared to Haskell). Considering this, it seems strange to me that `findIndex` (and three related functions) have the type
```purescript
findIndex :: forall a. (a -> Boolean) -> [a] -> Number
```
and fail with `-1`. While this is technically not 'unsafe' behavior, the `-1` should not be used as an array index. Right now, I don't see how the `-1` value is useful other than signalling failure. For me, it would feel more natural if the type would simply be
```purescript
findIndex :: forall a. (a -> Boolean) -> [a] -> Maybe Number
```
and failure would be indicated by `Nothing`.

garyb:
I'm still finishing up some basic assertion-style tests for everything, but here's the update for `Int`, and also resolves #36.


paf31:
All looks good to me. I expect the damage from this is pretty considerable, but we should definitely do it.

garyb:
For purescript/purescript#1038

PR #39, opened 2015-04-30
Add `cons`

hdgarrood:
I haven't been able to test this, but it was copied verbatim from `purescript-prelude`, so I think it probably works.

Refs purescript/purescript#1089


paf31:
:+1: Thanks!

PR #40, opened 2015-05-01
0.7 updates, WIP

paf31:
I probably still have work to do here.

/cc @garyb


garyb:
Are we not going to switch things around so `arrays` can depend on `foldable-traversable`? I'd like to avoid exposing the array specialised folds if we can. :)

Issue #41, opened 2015-05-02
concat - RangeError: Maximum call stack size exceeded


ethul:
On my machine I am getting a `RangeError` when using `concat` on an array with 150,000 elements; e.g.,

```javascript
function concat (xss) {
  var result = [];
  for (var i = 0, l = xss.length; i < l; i++) {
    result.push.apply(result, xss[i]);
  }
  return result;
}

var as = new Array(150000);
var bs = concat([as]);
```

```bash
$ node index.js

/index.js:4
    result.push.apply(result, xss[i]);
                ^
RangeError: Maximum call stack size exceeded
```

ethul:
Just saw that PR #24 also fixes this.

garyb:
Resolved by #42, even though it wasn't actually merged :smile:

PR #42, opened 2015-05-02
Resolving RangeError for concat

ethul:
Refactors the concat implementation to avoid the apply function due to a
RangeError for large arrays.

Resolves purescript/purescript-arrays#41


garyb:
I'm not sure when it happened, but we adopted this implementation at some point in the updates for 0.7. Thanks!

ethul:
Great! Glad it is resolved.

PR #43, opened 2015-05-18
Remove orphan Foldable/Traversable instances

hdgarrood:
This PR also adds .jscsrc and .jshintrc (copied from purescript-st). I added them with `git add --force`, so that the .gitignore didn't need to be modified.


garyb:
Oops, I guess I duplicated this effort in the end. Thanks anyway!

PR #45, opened 2015-07-12
Add example to uncons documentation

garyb:
If this looks good I'll add the example to the migration guide too.


paf31:
:+1: Thanks!

PR #52, opened 2015-12-09
Improve docs

hdgarrood:
See the commit messages for a description of the changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
 sudo: false
 node_js:
-  - 4.2
-  - 5.2
+  - 4
+  - 5
 env:
   - PATH=$HOME/purescript:$PATH
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: node_js
 sudo: false
 node_js:
-  - 0.10
+  - 4.2
+  - 5.2
 env:
   - PATH=$HOME/purescript:$PATH
 install:
--- a/docs/Data/Array.md
+++ b/docs/Data/Array.md
@@ -8,11 +8,35 @@ use cases. This module is useful when integrating with JavaScript libraries
 which use arrays, but immutable arrays are not a practical data structure
 for many use cases due to their poor asymptotics.
 
+In addition to the functions in this module, Arrays have a number of
+useful instances:
+
+* `Functor`, which provides `map :: forall a b. (a -> b) -> Array a ->
+  Array b`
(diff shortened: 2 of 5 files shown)


hdgarrood:
I thought it would be good to add details of the instances because a) `Array` is a native type so instances aren't listed, and b) this is probably confusing beginners: "Why is there no function for concatenating arrays in `Data.Array`?"

.travis.yml
```diff
@@ -1,7 +1,8 @@
 language: node_js
 sudo: false
 node_js:
-  - 0.10
+  - 4.2
+  - 5.2
```
garyb: Is it worth just specifying `4` and `5` here so the latest point release of each is used?

hdgarrood: I wasn't aware you could do that! Definitely sounds sensible. Will do.

docs/Data/Array.md
```diff
@@ -8,12 +8,36 @@ use cases. This module is useful when integrating with JavaScript libraries
...
+  -> Array b` (this is the same as `concatMap`).
+* `Semigroup`, which provides `(<>) :: forall a. Array a -> Array a ->
+  Array a`, for concatenating arrays.
+* `Foldable`, which provides a slew of functions for *folding* (also known
+  as *reducing*) arrays down to one value. For example,
+  `Data.Foldable.any` tests whether an array of `Boolean` values contains
+  at least one `true`.
+* `Traversable`, which provides the PureScript version of a for-loop,
+  allowing you to iterate over an array and accumulate effects.
+
```
garyb: Sorry, had another thought. Perhaps we could make these link to the appropriate classes rather than having to include types, etc? Or maybe both if you think it's useful to include this info here as well.

hdgarrood: That would be nice, but we'd have to choose a specific version of prelude/foldable-traversable to link to, which might cause problems if things change. These particular classes probably won't change that much, but other stuff in those packages might.

I had beginners in mind in particular with this paragraph, or rather, people who aren't yet familiar with all of these type classes. For example, I've encountered people more than once in IRC who were confused about why `Data.Array` doesn't have a function analogous to JS' `Array.prototype.concat`. So that's why I included this info here.

garyb: Ah, ok. I kinda assumed you could leave the version out of the URL and it would redirect to the latest, the way it does for the package route. But that wouldn't help anyway if things moved around.

Makes sense.

hdgarrood: I think giving people a way of linking to a particular type in the documentation could be really nice, but I haven't quite figured out how to do it well just yet.

PR #59, opened 2016-02-26
Add `partition` to Data.Array

raichoo:
I needed a `partition` function for Arrays so I made one. Pointers welcome, this currently feels like the straightforward way to implement it.

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -64,6 +64,7 @@ module Data.Array
   , concat
   , concatMap
   , filter
+  , partition
   , filterM
   , mapMaybe
   , catMaybes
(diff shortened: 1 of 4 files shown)


raichoo:
Interestingly enough this seems to break in a line that I didn't touch. I'm not familiar with the tools complaining here so I'm not sure if this is actually my mistake, it certainly looks like it isn't to me.

garyb:
Ugh, it's not you, jshint broke something a while ago. Can you change [this line](https://github.com/purescript/purescript-arrays/blob/f2ed8d23b1c9be70370698178e4bcf447c79bcc4/.jshintrc#L8) to `"strict": "global",` please? That should fix it.

I'm also amazed we don't already have `partition` :smile:

raichoo:
Good to know :) So was I ^^.

raichoo:
All fixed now.

src/Data/Array.purs
```diff
@@ -369,6 +370,13 @@ concatMap = flip bind
 -- | creating a new array.
 foreign import filter :: forall a. (a -> Boolean) -> Array a -> Array a
 
+-- | Partition an array using a predicate function, creating a set of
+-- | new arrays. One for the values satisfying the predicate function
+-- | and one for values that don't.
+foreign import partition :: forall a. (a -> Boolean)
+                         -> Array a
+                         -> { yes :: Array a, no :: Array a }
```
paf31: I like the return type, I frequently have to consult Hackage to remember the right way round in Haskell.

paf31:
:+1: LGTM, but could you please run `pulp docs`?

raichoo:
On its way. The output surprised me however. I looks like it touched more of the documentation than I actually changed.

paf31:
It looks like we forgot to build it last time something was merged.

Looks good, thanks!

sharkdp:
Looks good to me, too.

garyb:
Thanks @raichoo!

PR #26, opened 2015-02-11
Preserve Haskell semantics in take/drop

jacereda:
A possible fix for #25

PR #62, opened 2016-03-20
Treat negative arguments as zero, fixes #25

### LINKED ISSUE #25: Differences from Haskell in take/drop
Haskell:

    Prelude> take (-1) [1,2,3]
    []
    Prelude> drop (-1) [1,2,3]
    [1,2,3]

PureScript:

    > take (-1) [1,2,3]
      [1,2]
    
    > drop (-1) [1,2,3]
      [3]

I haven't seen the differences documented anywhere. If these are intended, a note in https://github.com/purescript/purescript/wiki/Differences-from-Haskell would be nice.

IMHO take/drop should keep haskell semantics and 'slice' should be exported keeping JS semantics.

sharkdp:
This PR changes the behavior of `take` for negative arguments to be consistent with that of `drop` (same behavior as in Haskell).
``` purs
take (-2) [1, 2, 3]) == []
```

This is similar to the fix in PR #26, which is outdated.

### DIFF
--- a/src/Data/Array.js
+++ b/src/Data/Array.js
@@ -243,6 +243,12 @@ exports.slice = function (s) {
   };
 };
 
+exports.take = function (n) {
+  return function (l) {
+    return n < 1 ? [] : l.slice(0, n);
+  };
+};
+
 exports.drop = function (n) {
   return function (l) {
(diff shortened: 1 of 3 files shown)


garyb:
:+1:

PR #63, opened 2016-03-20
Update documentation, resolves #56

### LINKED ISSUE #56: union, unionBy are not symmetrical in their arguments
This seems weird:
```purs
union [0, 0] [1, 1] == [0, 0, 1]
```
I would have expected the result to be `[0, 1]`, but at least it should be symmetrical in both arguments. I would consider several possible fixes:

1. Remove all duplicates (i.e. `union a b == nub (a <> b)`)
2. Keep the duplicates within both of the arguments (i.e. `union [0, 0] [1, 1] == [0, 0, 1, 1]`).
3. Remove `union` and `unionBy` from `Data.Array` (given that `Set` is the better data structure for something like this)
4. Document the behavior if it is intended

I would vote for 1 or 3.

sharkdp:
See the discussion in #56

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -519,14 +519,16 @@ nubBy eq xs =
     Just o -> o.head : nubBy eq (filter (\y -> not (o.head `eq` y)) o.tail)
     Nothing -> []
 
--- | Calculate the union of two lists.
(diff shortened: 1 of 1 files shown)


garyb:
:+1:

PR #66, opened 2016-05-23
Use ~> in types

garyb:
So, I'm not sure about this, but it'd be good to get some feedback.

Using `(~>)` wherever it can be used leads to this sort of thing:

``` purescript
head :: Array ~> Maybe
reverse :: Array ~> Array
```

Which is good in some ways, as it shows that these operations are "structural" and care not about the contents at all, however, it might be super confusing to the uninitiated?

I think I do prefer these though:

``` purescript
toUnfoldable :: forall f. Unfoldable f => Array ~> f
fromFoldable :: forall f. Foldable f => f ~> Array
```

/cc @paf31 @AppShipIt @hdgarrood anyone else who's interested


paf31:
I'll just mention one other possible issue - this might change type inference, since you move the `forall` to the right of an `->` in some cases. I'm not sure if this could be an issue or not.

src/Data/Array.purs
```diff
@@ -436,19 +435,19 @@ foreign import sortImpl :: forall a. (a -> a -> Int) -> Array a -> Array a
...
 
 -- | Keep only a number of elements from the start of an array, creating a new
 -- | array.
-foreign import take :: forall a. Int -> Array a -> Array a
+foreign import take :: Int -> Array ~> Array
 
 -- | Calculate the longest initial subarray for which all element satisfy the
 -- | specified predicate, creating a new array.
 takeWhile :: forall a. (a -> Boolean) -> Array a -> Array a
 takeWhile p xs = (span p xs).init
```
paf31: Unrelated, but I just noticed this involves allocating two arrays when we only return one (`init`).

garyb:
Do you think that would make a difference? In rank terms it only makes a difference if the `forall` is on the left of an `->`, right?

paf31:
It should be okay, but might result in some funny inferred types. `slice 1` would have inferred type `Int -> forall a. Array a -> Array a`, I think.

PR #68, opened 2016-07-21
Adds mapWithIndex (map with a generated index)

damncabbage:
`mapWithIndex :: forall a b. (Int -> a -> b) -> Array a -> Array b`

This PR came about as a result of asking in the Gitter channel to see if anyone else writes their own version of this function: https://gitter.im/purescript/purescript?at=5785c87959cfbd4c5e93113f

(CC: @dkoontz, @felixSchl)

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -68,6 +68,7 @@ module Data.Array
   , filterM
   , mapMaybe
   , catMaybes
+  , mapWithIndex
 
   , sort
   , sortBy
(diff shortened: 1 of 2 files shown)


damncabbage:
(This is already called `imap` in the [Haskell `vector` package](https://hackage.haskell.org/package/vector-0.11.0.0/docs/Data-Vector-Generic.html#v:imap), in case you're not happy with the name.)

garyb:
`mapWithIndex` would be my preferred name :+1: we use `imap` for invariant functors: https://pursuit.purescript.org/packages/purescript-invariant/1.0.0/docs/Data.Functor.Invariant#v:imap

paf31:
Looks good! Agreed on the name.

PR #72, opened 2016-10-06
Added the find function

Risto-Stevcev:


garyb:
`find` is in `Foldable` :smile:

Risto-Stevcev:
Ah! I didn't notice that. What do you think of re-exporting it from `Foldable` so that it's available in `Data.Array` also?
`purescript-foldable-traversable` is already a dependency for `purescript-arrays`

garyb:
Yeah, that makes sense. There are probably a few other common functions we could export from there too.

/cc @paf31

paf31:
👍

garyb:
Oh sorry, I see you pushed another change. I ended up including the re-exports in my other PR.

Risto-Stevcev:
Ah it's okay, I was being slow :sweat_smile:

Issue #71, opened 2016-10-04
Use an index instead of uncons to make array functions more efficient.


jutaro:
The implementation of many array functions, like filterM, span, groupBy, ... uses uncons for implementation. It would be more efficient to use an index and has no known disadvantages (other then maybe some subjective aesthetic property).

If we can agree on this I can contribute code, but I don't want to work on this if it is not wanted.

paf31:
Let's list out the functions which have possible issues, since I think each could have a slightly different fix.

`filterM` - I'm not sure how you could make this more efficient for a general `Monad`.

`span` and `groupBy` could potentially make use of `STArray`. Something like `discrimination` is probably worth investigating for sorting and grouping too.

hdgarrood:
I think converting to a list before doing `filterM` and then converting back to an Array at the end would be better than the current implementation; right now it's at least O(n^2) I think.

hdgarrood:
Mostly a note to myself: now, it's just `groupBy`, `nubBy`, `unzip`, and `foldM` that use `uncons` (and perhaps shouldn't).

hdgarrood:
`groupBy` is now done, so I'll make a checklist for the remaining ones:

* [ ] `nubBy`
* [ ] `unzip`
* [ ] `foldM`

hdgarrood:
That sounds good to me, provided that the `Foldable` has better performance than the current `foldM` — which is almost certainly true but probably still worth checking, at least for one or two monads.

hdgarrood:
`nubByEq` was the last problematic one, so I'm going to close this, but of course please continue to open issues if you find places where performance could be improved.

PR #73, opened 2016-10-07
Prepare for 2.0 release

### LINKED ISSUE #70: foldM is not tail recursive and crashes on large collections
E.g. this implementation does not crash and is more efficient, but it has an additional type class constraint. I'm not aware of the implications of this constraint

```purescript
foldM :: forall m a b. MonadRec m => (a -> b -> m a) -> a -> Array b -> m a
foldM func init array = tailRecM2 go init 0
  where
    go res index | index >= length array = pure (Right res)
    go res index                         = do
        res' <- func res (unsafePartial (unsafeIndex array index))
        pure $ Left {a:res', b:(index + 1)}
```

garyb:
This re-exports the more common stuff out of `Foldable` and `Traversable`. Anything else we want to include?

Resolves #72, resolves #70

As for #71, I guess we can do that post-release, since it shouldn't affect the interface.


paf31:
> do we want to limit foldM to MonadRec?

I would say people should use `FreeT` to make things stack-safe first, but there is a valid use case where that might be overkill: using `foldM` with `List` to generate a graph. Indeed, I use that in the book. So I think yes, let's add this as a separate function for now.

garyb:
There's also the point raised in https://github.com/purescript/purescript/issues/2359 about the not insignificant overhead if you don't need it too. I'll add it as a separate thing then. `foldMRec` good for the name?

garyb:
I went with `foldRecM` to match `tailRecM`, etc.

PR #75, opened 2016-10-18
Add unsnoc

joshuahhh:
This is based off the v1.1.0 tag, since the newer versions require newer versions of Purescript... I'll be happy to rebase when things settle down, or anyone else can feel free to do so.

PTAL – thanks!

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -49,6 +49,7 @@ module Data.Array
   , tail
   , init
   , uncons
+  , unsnoc
 
   , (!!), index
   , elemIndex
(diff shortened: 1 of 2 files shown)


paf31:
Looks good, thanks!

Do you need this now for 0.9.3? I think we could probably make a v1.2.0 tag off of v1.1.0 (@garyb?) if so. If it's not urgent, I think let's rebase this on top of `master`, since it should be easier to upgrade soon.

joshuahhh:
Not urgent at all. I just didn't want to mess around with 0.10.1 yet (to run tests for this).

I'll rebase off master and see if the Travis build gives me the confidence I need!

hdgarrood:
What do people think about using `import Data.Array as A` in the tests instead of an explicit import list? That could help prevent conflicts like this one.

garyb:
Yeah, sounds good to me.

hdgarrood:
I merged your commit in #75. Thanks!

PR #81, opened 2016-11-20
Speed up `difference`

hdgarrood:
Also add complexity to docs.

difference

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -573,6 +573,8 @@ unionBy eq xs ys = xs <> foldl (flip (deleteBy eq)) (nubBy eq ys) xs
 
 -- | Delete the first element of an array which is equal to the specified value,
 -- | creating a new array.
+-- |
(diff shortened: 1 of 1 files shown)


garyb:
Love it 👍

hdgarrood:
This is the code I was using in that benchmark, by the way:

```purescript
benchDifference :: Benchmark
benchDifference = mkBenchmark
  { slug: "difference"
  , title: "Difference"
  , sizes: A.range 1 5 <#> (_ * 1000)
  , sizeInterpretation: "Length of the array"
  , inputsPerSize: 1
  , gen: \n -> map prepare (vectorOf n (arbitrary :: Gen Int))
  , functions: [ benchFn "current"   (uncurry currentDifference)
               , benchFn "suggested" (uncurry suggestedDifference)
               ]
  }

  where
  prepare xs =
    Tuple xs (A.reverse xs <> [1,2,3])
```

PR #83, opened 2016-11-20
Add filterA, a faster version of filterM

hdgarrood:
Change filterM to just be filterA in order to have this be a non-breaking change; I was just thinking it would be nice for all these performance improvements to be available for people without having to update to the next major version.

This removes another use of uncons, refs #71.

The new devDependency on `Const` is because that seemed like the best choice for testing that effects are applied in the right order; the tests were not previously testing that `filterM` applies effects in the right order.

Benchmarking code:

```purescript
benchFilterM :: Benchmark
benchFilterM = mkBenchmark
  { slug: "filterM"
  , title: "filterM"
  , sizes: A.range 1 5 <#> (_ * 500)
  , sizeInterpretation: "Length of the array"
  , inputsPerSize: 1
  , gen: \n -> vectorOf n (arbitrary :: Gen Int)
  , functions: [ benchFn "current"   (currentFilterM pred)
               , benchFn "suggested" (suggestedFilterM pred)
               ]
  }
  where
  pred :: Int -> WriterT Unit Identity Boolean
  pred x = tell unit $> even x
```

I had to stop at 2500 elements since the current implementation stack overflows at around 3k, but the new implementation should work for much larger arrays, because of the stack-safety of the `Traversable Array` instance. I tested this implementation with up to 10k elements and it seemed to work fine.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -26,6 +26,7 @@
   },
   "devDependencies": {
     "purescript-assert": "^2.0.0",
-    "purescript-console": "^2.0.0"
+    "purescript-console": "^2.0.0",
+    "purescript-const": "^2.0.0"
   }
 }
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -67,6 +67,7 @@ module Data.Array
   , concatMap
   , filter
   , partition
+  , filterA
   , filterM
   , mapMaybe
   , catMaybes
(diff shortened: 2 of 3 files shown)


hdgarrood:
Oh, nearly forgot - benchmark results!
filterm

hdgarrood:
Also I should probably check - are we ok with deprecating `filterM` now that we are exporting `filterA`?

paf31:
👍 LGTM, thanks (and also +1 to deprecating `filterM`)

hdgarrood:
Great. :)

Issue #78, opened 2016-11-08
Return more than one STArray from an ST computation?


hdgarrood:
Given the type of `runSTArray`:

```purescript
runSTArray ::
  forall a r.
    (forall h. Eff (st :: ST h | r) (STArray h a)) -> Eff r (Array a)
```

it seems to me that it's a nice alternative to `freeze` + `runST`, which would involve an unnecessary copy, but I can't see a way to take advantage of this if I want to return more than one STArray from the computation. This would be helpful, for example, for efficiently implementing functions like `unzip`. Can anyone think of a good way of addressing this?

hdgarrood:
I just thought of something - what if we also exported this from `Data.Array.ST`:

```purescript
runSTArray' ::
  forall f a r.
    (forall h. Eff (st :: ST h | r) (f (STArray h) a))
    -> Eff r (f Array a)
runSTArray' =
  unsafeCoerce
```

Then you could write:

```purescript
newtype ArrayPair arr a = ArrayPair (Tuple (arr a) (arr a))
newtype GroupedArray arr a = GroupedArray (arr (NonEmpty arr a))
```

and then you could use these types for `f` to implement `unzip` and `groupBy` respectively with no unnecessary copying. Does that sound sensible?

paf31:
A nitpick: I think you can simplify things a bit by getting rid of the `a` since it's not mentioned elsewhere:

```haskell
runSTArray' ::
  forall f a r.
    (forall h. Eff (st :: ST h | r) (f (STArray h)))
    -> Eff r (f Array)
runSTArray' = unsafeCoerce

newtype ArrayPair a arr = ArrayPair (Tuple (arr a) (arr a))
newtype GroupedArray a arr = GroupedArray (arr (NonEmpty arr a))
```

Malicious users could sidestep `runSTArray'` by defining some `f` using the FFI potentially (by essentially emulating a type family), but I think that's probably ok.

hdgarrood:
Great, thanks. This is what I ended up doing in the groupBy PR, as it happens. :)

PR #77, opened 2016-11-07
Remove a couple of uses of uncons

hdgarrood:
Fixes #76, partially addresses #71.

I benchmarked this toUnfoldable vs the current one and my suggested one seems to win fairly convincingly:

tounfoldable 1

tounfoldable

PR #84, opened 2016-11-21
Iterator groupby

hdgarrood:
I recommend reviewing these commits one at a time, as they are each independent of each other. The `runSTArray'` commit is more or less what I was describing in #78.

### DIFF
--- a/src/Data/Array/ST.purs
+++ b/src/Data/Array/ST.purs
@@ -18,7 +18,7 @@ module Data.Array.ST
   ) where
 
 import Prelude
-import Control.Monad.Eff (Eff)
+import Control.Monad.Eff (Eff, Pure, runPure)
 import Control.Monad.ST (ST)
 import Data.Maybe (Maybe(..))
 import Data.Newtype (class Newtype, unwrap)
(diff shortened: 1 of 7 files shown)


hdgarrood:
Benchmark results:

groupby

I was honestly expecting a slightly more drastic difference, although it's perhaps worth noting that the implementation labelled "current" is what's in master right now, not what's in the most recently released version. However, the version of `groupBy` in `master` at the moment uses `span` in its implementation, which had a very nice perf boost in #77, so I expect the difference that people will actually see between `groupBy` in their code now and the next time they upgrade will be quite a bit larger than it looks in this graph.

src/Data/Array/ST.purs
```diff
@@ -33,14 +36,32 @@ foreign import data STArray :: * -> * -> *
...
+-- | arrays. Use this function as you would use `runST` to freeze a mutable
+-- | reference.
+-- |
+-- | The rank-2 type prevents the reference from escaping the scope of `runSTArray'`.
+runSTArray'
+  :: forall f r
+   . (forall h. Eff (st :: ST h | r) (f (STArray h)))
+  -> Eff r (f Array)
+runSTArray' =
+  unsafeCoerce
```
paf31: I'm not convinced it's safe to just coerce here, without copying the arrays. For example, what about something like:

```purescript
runSTArray do 
  xs <- initSomeSTArray
  setTimeout 1000 (modifySomehow xs)
  pure xs
```

The return value is supposed to be a regular array, but it will get modified in place one second after the function returns.

hdgarrood: Oh, good point. I guess the current `runSTArray` function has the same issue too? What if we restrict the row type to a closed row?

paf31: Yes, I think restricting to a closed row should solve the problem.

hdgarrood: Cool, thanks. In that case I think I will publish a minor release before doing that, since this will be a breaking change.

hdgarrood:
I decided to change `runSTArray` to just return an `Array a` instead of returning a `Pure (Array a)`, since now we know it's going to be a `Pure` so we might as well use `runPure` on it.

paf31:
Thanks again, and sorry for the delay!

paf31:
Hmm, I just remembered this is a breaking change. What shall we do here? Revert, or make a major release?

hdgarrood:
No problem! Up to you whether you think it's worth merging now and making a major release. We could also maybe make a branch off of master from before any breaking changes were merged in case we ever want to do a minor or patch release before the next major one?

paf31:
I think we have a few libraries which need major releases:

- `arrays`
- `foldable-traversable`
- `strings`

Any others?

hdgarrood:
I'm not aware of any off the top of my head

PR #85, opened 2016-11-24
[BREAKING] Remove filterM

hdgarrood:


rightfold:
Why?

hdgarrood:
Because we have `filterA` now which only needs an `Applicative`, and is significantly faster - both asymptotically and in practice.

hdgarrood:
Oh actually I guess the perf argument doesn't really make sense because filterM is now just an alias for filterA. I think filterA should be preferred because it can be misleading to ask for a Monad when you're only using an Applicative.

rightfold:
Maybe time to try out the shiny new `Warn` constraints?

garyb:
That'd be breaking too :smile:

PR #87, opened 2016-12-24
[BREAKING] Iterator groupby

paf31:
Reopening #84 after my mistake (this requires a major release)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -22,7 +22,8 @@
     "purescript-st": "^2.0.0",
     "purescript-tailrec": "^2.0.0",
     "purescript-tuples": "^3.0.0",
-    "purescript-unfoldable": "^2.0.0"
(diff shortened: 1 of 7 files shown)


hdgarrood:
Maybe we can do this in a non-breaking way by marking `runSTArray` as deprecated and also exporting a new function `unsafeFreeze :: forall h a eff. STArray h a -> Eff (st :: ST h | eff) (Array a)` which requires the caller to ensure that the original `STArray` is not modified afterwards? That would also make the API (as well as code that uses it) a bit simpler, I think: e.g. you no longer need to define a newtype to implement things like `groupBy`.

hdgarrood:
I'm going to try this in a separate branch and see how much I like it.

hdgarrood:
Resolved by #89.

PR #88, opened 2017-01-18
Use ~>

mlang:


hdgarrood:
I think we discussed this before somewhere, would you mind having a quick search through the closed issues/PRs in this repo?

garyb:
I did this one: #66, but applied it more widely, and wasn't sure if it would end up more confusing in the end. I think it's probably fine for these though.

hdgarrood:
Ok, cool. Let's merge this then?

PR #90, opened 2017-01-20
Added zipWith3.

michal-jewel:
I can also add further zipWith functions for convenience.


michal-jewel:
(And efficiency.) It would be also good to move more into Array-based functions, not just lists.

hdgarrood:
I think the ST stuff is advanced enough now that we should avoid defining any more of these kinds of functions with the FFI, and in the long term aim to eliminate FFI use completely in Data.Array, even.

michal-jewel:
But will ST stuff give equally understandable generated code, as use of zipWith3 which is implemented as a single loop without additional calls?

hdgarrood:
With the Eff-specific optimisations, likely yes.

hdgarrood:
But even if not I'd rather do it with ST for now anyway and work out how to better generate/optimise that code later.

michal-jewel:
I have trouble reconciling your comment with that of paf31, who indicated the need to add native functions to get within 15% difference to native JavaScript. Is it not a priority any more? Or did the PureScript compiler get inlining yet?

hdgarrood:
Where was that?

michal-jewel:
https://github.com/purescript/purescript/issues/326

hdgarrood:
Ah, right. My impression is that the compiler is a fair bit cleverer now than when that discussion took place, and also it looks like `STArray` didn't exist back then. Now that alternate backends are a thing it makes sense to avoid the FFI, because the more code there is that uses the FFI, the more effort it is to maintain alternate backends.

Performance is still a priority, but if we drop down to the FFI every time we want a function like this, we'll never get any closer to being able to implement a similarly efficient version of functions like this within the language (which should be a long term goal IMO).

There's inlining now, but only for a specific hardcoded set of functions. A more general inliner is planned but hasn't been implemented yet.

michal-jewel:
I am interested in this issue, since I have a performance problem now, that was solved by using inlined zipWith3. Can you point me to competitive implementation using STArray?

hdgarrood:
No, I don't have one to hand. `STArray` algorithms generally look fairly similar to the JS equivalents, though, so I don't think it would be particularly difficult to do.

If you're happy with this implementation of `zipWith3`, you're more than welcome to use it within your project, but I think we should be more careful about FFI use in the core libraries.

hdgarrood:
It might also be possible to implement a fast version of this without the FFI by using the `Unfoldable Array` instance, come to think of it.

michal-jewel:
I thought this was clear implementation, and am surprised that it is rejected without even a request of a benchmark.

It may be that I was wrong thinking that core libraries are a place to share all basic functions that are useful to everyone. I guess a spirit of sharing means different things for different people.

hdgarrood:
I agree that it is a very clear implementation, and it does look like it would be very fast too, but as I've tried to explain there are other criteria that we need to worry about here.

The core libraries are indeed a place to share basic functions that are useful to everyone, but if you want your changes to be merged, you have to be prepared to listen to and act on feedback.

michal-jewel:
I understand your concern about increasing the maintenance load. But of course, the fastest growing open source projects tend to react a bit differently.

paf31:
> But of course, the fastest growing open source projects tend to react a bit differently.

Fast growth is not the measure we're trying to optimize for here.

In future, opening an issue before making a PR is a good way to avoid this sort of situation.

My 2¢: I'd be fine with defining `zipWith3` in terms of `zipWith2`. If/when we get inlining, it should be possible to optimize that well.

PR #93, opened 2017-02-24
Faster concat.

dikmax:
Simply calling `concat` is much faster that going through loop.

### DIFF
--- a/src/Data/Array.js
+++ b/src/Data/Array.js
@@ -188,14 +188,7 @@ exports.reverse = function (l) {
 };
 
 exports.concat = function (xss) {
-  var result = [];
-  for (var i = 0, l = xss.length; i < l; i++) {
-    var xs = xss[i];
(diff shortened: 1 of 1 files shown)


garyb:
Unfortunately this version isn't "safe" - see #42

hdgarrood:
I'm also not sure that this implementation is faster anyway, at least I don't think it was last time i checked. Did you do any benchmarks?

dikmax:
Here's some benchmarks <https://jsperf.com/concat-vs-push100>

I'll try to think something that wouldn't cause rangeerror.

hdgarrood:
I think benchmarks need to use a variety of array sizes to be useful, we can't see the whole picture from a test with just one input. I think it's a bit of a shame that jsperf encourages benchmarks like this, actually. I'd prefer to see a benchmark using a variety of inputs with different sizes before making a change like this.

You might find my library purescript-benchotron useful for creating a more comprehensive benchmark.

PR #94, opened 2017-02-25
Faster concat.

dikmax:
New attempt. I've performed different benchmarks and appears that `Array.prototype.concat.apply` is faster than a loop. But it's unsafe to call on big arrays because it crashes with stack overflow.  That's why I propose to use the fast version on smaller arrays, and old on bigger.

Size limit (10000) is chosen to be safe. I didn't check exact limits, but Chrome and Firefox are ok with 10x this limit, whereas Safari only with 5x.

Here are some benchmark images. On all images `default` corresponds to current `concat`.

Smaller arrays 10x10 to 200x200.

concat2

Bigger arrays 100x100 to 1000x1000.

concat

Small array (1..10) of big arrays with 100000 items.

concat-small

Next two are fallback to old methods. That's why results are same.

Big array with (100000..1000000) of small arrays of 1.

concat-large

Big array with (100000..1000000) small arrays of 10.

concat-large2

And here's a link to benchmark itself: <https://github.com/dikmax/purescript-array-benchmarks>.

### DIFF
--- a/src/Data/Array.js
+++ b/src/Data/Array.js
@@ -188,6 +188,12 @@ exports.reverse = function (l) {
 };
 
 exports.concat = function (xss) {
+  if (xss.length <= 10000) {
+    // This method is faster, but it crashes on big arrays.
+    // So we use it when can and fallback to simple variant otherwise.
+    return Array.prototype.concat.apply([], xss);
+  }
+
   var result = [];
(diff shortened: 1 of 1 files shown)


hdgarrood:
Looks good to me, although could you please add a comment to explain what's going on there?

PR #95, opened 2017-02-26
Faster replicate.

dikmax:
Faster version of replicate. Here are benchmark results.

replicate

You can check benchmark itself here: <https://github.com/dikmax/purescript-array-benchmarks>.

### DIFF
--- a/src/Data/Array.js
+++ b/src/Data/Array.js
@@ -16,7 +16,17 @@ exports.range = function (start) {
   };
 };
 
-exports.replicate = function (count) {
+var replicate = function (count) {
+  return function (value) {
+    if (count < 1) {
+      return [];
+    }
+    var result = new Array(count);
+    return result.fill(value);
+  };
+};
+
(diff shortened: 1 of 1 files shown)


hdgarrood:
This looks great but sadly I don't think we can justify using this implementation due to its relative lack of support: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/fill

What does it look like for smaller arrays, where the argument is between, say, 1 and 10000?

dikmax:
Here are smaller arrays.

replicate 1

I can write version, which selects method depending on the presence of `fill` in arrays prototype. What do you think?

hdgarrood:
Great! I think that should work, yeah. It would be good to have a version that only selects the method once, i.e. ideally we would only check once for `Array.prototype.fill`, not every time `replicate` is called, I think.

dikmax:
Fixed.

PR #96, opened 2017-03-16
partial functions for ST arrays

matthewleon:
I was surprised to see that partial functions exist for `Array` but not `STArray`, given the performance-sensitive contexts in which `STArray` is used. I've used these functions in `purescript-mersenne` (https://github.com/matthewleon/purescript-mersenne) to significantly increase performance in a tight loop.


paf31:
Thanks!

PR #97, opened 2017-03-21
Haskell inspired sortWith

negator:
Support for the GHC Ext `sortWith` function, which accepts a projection function.

http://hackage.haskell.org/package/base-4.9.1.0/docs/GHC-Exts.html#v:sortWith

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -75,7 +75,7 @@ module Data.Array
 
   , sort
   , sortBy
-
+  , sortWith
   , slice
   , take
   , takeWhile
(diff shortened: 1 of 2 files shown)


src/Data/Array.purs
```diff
@@ -469,6 +469,13 @@ sortBy comp xs = sortImpl comp' xs
     EQ -> 0
     LT -> -1
 
+-- | Sort the elements of an array in increasing order, where elements are
+-- | sorted based on a projection
+sortWith :: forall a b. Ord b => (a -> b) -> Array a -> Array a
+sortWith f xs = sortBy comp' xs
```
paf31: You can use `sortWith f = sortBy (comparing f)`

negator: Excellent 👍

PR #101, opened 2017-05-15
batch update and modify functions

matthewleon:
Also exports ST utility functions `modifySTArray` and `withArray`.

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -71,6 +71,8 @@ module Data.Array
   , mapMaybe
   , catMaybes
   , mapWithIndex
+  , updateAtIndices
+  , modifyAtIndices
 
   , sort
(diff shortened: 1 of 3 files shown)


src/Data/Array.purs
```diff
@@ -71,6 +71,8 @@ module Data.Array
   , mapMaybe
   , catMaybes
   , mapWithIndex
+  , updateAtIndices
```
hdgarrood: Perhaps put these next to `updateAt` and `modifyAt` in the export list, so that when we (I) eventually get around to ordering declarations in HTML docs based on their order of appearance in the export list the docs read better?

matthewleon: Done.

paf31:
Thanks again!

PR #103, opened 2017-06-13
Array.fromIndexation

matthewleon:
A single function to produce an `Array` from an `Array.ST.Iterator` indexation function.

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -30,6 +30,7 @@
 module Data.Array
   ( fromFoldable
   , toUnfoldable
+  , fromIndexation
   , singleton
   , (..), range
   , replicate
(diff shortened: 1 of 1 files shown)


hdgarrood:
I'm not really a fan of this word "indexation", is it used elsewhere?

paf31:
What about `fromIndexed`?

Is there a use case for this?

matthewleon:
> I'm not really a fan of this word "indexation"

I'm not a big fan either. I just couldn't come up with anything better on the spot. I guess `fromIndexed` is better.

> Is there a use case for this?

Use case is the same as the tools in `Data.Array.ST.Iterator`. This just seemed to be a missing piece of that puzzle. Admittedly, my own use case is in benchmarking (https://github.com/matthewleon/purescript-foldablerec/blob/master/bench/Bench/Main.purs#L68), but that is a bit contrived. To me the most immediate use case would be for interop with a JS library where you have indexed data for which you could easily write an `Int -> Maybe a` indexation function. Perhaps this should be documented that way.

I'm not particularly attached to this PR. I was just surprised that `Data.Array.ST.Iterator` existed without a convenient "bridge" into `Data.Array`.

hdgarrood:
I'm not sure I understand the use case here. The purpose I had in mind when I created that module was more iterating over arrays you've already been given when you're working in an `ST` computation, so the `Array` -> `Iterator` direction is the direction I'm more interested in.

The reason I didn't provide a function like this originally is that I was imagining that the argument to `iterator` would almost always be of the form `(arr !! _)`, where `arr` is an array, and in that case this new function wouldn't do anything.

Can you give a concrete example of where you might get a function `Int -> Maybe a` from other than from indexing into an array, and why you might want to use such a function to create an array in this way?

matthewleon:
> The purpose I had in mind when I created that module was more iterating over arrays you've already been given when you're working in an ST computation, so the Array -> Iterator direction is the direction I'm more interested in.

Understood. Closing this.

PR #105, opened 2017-06-20
Use a doc-comment for `withArray`

hdgarrood:


paf31:
👍

hdgarrood:
It would be nice to add an example for this function too, but I wasn't able to immediately think of one. Come to think of it I think many of the functions in this module would benefit from examples. Hopefully I'll get around to that eventually..!

PR #106, opened 2017-06-20
Update npm dependencies

hdgarrood:
This means that the npm scripts for this project no longer rely on
wrapper scripts, but use the `purs` executable directly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,9 @@
     "test": "pulp test"
   },
   "devDependencies": {
-    "eslint": "^3.17.1",
-    "pulp": "^10.0.4",
-    "purescript-psa": "^0.5.0-rc.1",
+    "eslint": "^4.0.0",
+    "pulp": "^11.0.0",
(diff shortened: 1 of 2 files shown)


garyb:
They're not using the wrapper scripts already - it currently works because `purescript-psa` was updated. Fixing that fixed pulp before pulp itself updated. :wink:

Could bump that to the non-RC release at the same time here I guess? Or just `ncu -a` for that matter.

hdgarrood:
Ohhh, I see :)

`ncu -a` also updates eslint to 4.0.0. I'll just browse through the eslint changelog quickly to see if anything has changed which could require us to do something.

hdgarrood:
Yep, nothing in eslint 4.0.0 seems to be a problem.

matthewleon:
Merge this? Maybe with more-up-to-date deps if that makes sense?

garyb:
I've merged a more updated version of this in the 0.12 branch. 🙂

Issue #110, opened 2017-06-22
Runtime error when using partial ST functions


mhuisi:
Using peekSTArray and pokeSTArray from Data.Array.ST.Partial yields a runtime error:
```TypeError: Cannot read property '[object Array]' of undefined```
The generated JavaScript for peekSTArray looks like this:
```
exports.peekSTArray = function (xs) {
  return function (i) {
    return function () {
      return xs[i];
    };
  };
};
```
But is called like this:
```Data_Array_ST_Partial.peekSTArray(dictPartial)(a)(i)();```
I.e. xs[i] ends up as dictPartial[a].
I suspect this is because of peekSTArray being imported like:
```
foreign import peekSTArray
  :: forall a h r
   . Partial
  => STArray h a
  -> Int
  -> Eff (st :: ST h | r) a
```
instead of
```
peekSTArray
  :: forall a h r
   . Partial
  => STArray h a
  -> Int
  -> Eff (st :: ST h | r) a
peekSTArray = peekSTArrayImpl

foreign import peekSTArrayImpl
  :: forall a h r
   . STArray h a
  -> Int
  -> Eff (st :: ST h | r) a
```
which is used instead in some places, for instance unsafeIndex from Data.Array.
The same is true for pokeSTArray.

I apologize if this is not actually an issue; I am very new to the FP world and only started learning PureScript a few days ago.

natefaubion:
This is definitely an issue. The foreign import specifies a `Partial` constraint, but the implementation does not account for it. Your recommendation looks correct.

hdgarrood:
Seconded. Thanks for the report!

mhuisi:
Fixed with b363605.

PR #111, opened 2017-06-22
prevent runtime error for partial functions for ST arrays

mhuisi:
See #110.


paf31:
Thanks! Could you please add a small test so we avoid this in future?

paf31:
Great, thanks very much!

PR #116, opened 2017-09-22
Preallocate result array for range

jacereda:
For `range 0 1000`:

No prealloc:
```
   range/1000           6020.450 ns 0.998 r2       166100.536 ops/s 

```
Prealloc:

```
   range/1000           3720.389 ns 0.995 r2       268789.072 ops/s 
```

### DIFF
--- a/src/Data/Array.js
+++ b/src/Data/Array.js
@@ -7,7 +7,7 @@
 exports.range = function (start) {
   return function (end) {
     var step = start > end ? -1 : 1;
-    var result = [];
+    var result = new Array(step * (end - start));
     var i = start, n = 0;
     while (i !== end) {
       result[n++] = i;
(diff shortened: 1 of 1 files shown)


jacereda:
Ups, I always forget `range` is inclusive...

jacereda:
With the correct size:

```
   range/1000           2580.780 ns 0.998 r2       387479.709 ops/s 
```

src/Data/Array.js
```diff
@@ -7,7 +7,7 @@
 exports.range = function (start) {
   return function (end) {
     var step = start > end ? -1 : 1;
-    var result = [];
+    var result = new Array(step * (end - start) + 1);
```
paf31: This seems wrong. Isn't this too large when `step > 1`?

jacereda: What do you mean? It can only be -1 or 1...

paf31: Oops, sorry, I misread :man_facepalming:

paf31:
Thanks!

PR #121, opened 2017-12-15
added examples to array module documentation

csicar:
this commit adds simple examples to the documentation comments for most functions defined in `Data.Array`

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -628,7 +628,7 @@ catMaybes = mapMaybe id
 -- | with the new elements.
 -- |
 -- | ```purescript
--- | mapWithIndex (\index element -> (show index) <> element) ["Hello", "World"] = ["0Hello", "1World"]
+-- | mapWithIndex (\index element -> show index <> element) ["Hello", "World"] = ["0Hello", "1World"]
 -- | ```
 -- |
 mapWithIndex :: forall a b. (Int -> a -> b) -> Array a -> Array b
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -604,10 +604,10 @@ filterA p =
 -- | which contain a value, creating a new array.
 -- |
 -- | ```purescript
--- | parseEmail :: String → Maybe Email
+-- | parseEmail :: String -> Maybe Email
 -- | parseEmail = ...
 -- |
--- | mapMaybe parseEmail ["a.com", "email@example.com", "--"] = [Email {user: "email", domain: "example.com"}]
+-- | mapMaybe parseEmail ["a.com", "hello@example.com", "--"] = [Email {user: "hello", domain: "example.com"}]
 -- | ```
 -- |
 mapMaybe :: forall a b. (a -> Maybe b) -> Array a -> Array b
@@ -628,7 +628,7 @@ catMaybes = mapMaybe id
 -- | with the new elements.
 -- |
 -- | ```purescript
--- | mapWithIndex (\index element → (show index) <> element) ["Hello", "World"] = ["0Hello", "1World"]
+-- | mapWithIndex (\index element -> (show index) <> element) ["Hello", "World"] = ["0Hello", "1World"]
 -- | ```
 -- |
 mapWithIndex :: forall a b. (Int -> a -> b) -> Array a -> Array b
@@ -844,7 +844,7 @@ group' = group <<< sort
 -- | specified equivalence relation to detemine equality.
 -- |
 -- | ```purescript
--- | groupBy (\a b → odd a && odd b) [1, 3, 2, 4, 3, 3] = [[1], [3], [2, 4], [3], [3]]
+-- | groupBy (\a b -> odd a && odd b) [1, 3, 2, 4, 3, 3] = [[1], [3], [2, 4], [3], [3]]
 -- | ```
 -- |
 groupBy :: forall a. (a -> a -> Boolean) -> Array a -> Array (NonEmpty Array a)
@@ -862,7 +862,7 @@ groupBy op xs =
 -- | Remove the duplicates from an array, creating a new array.
 -- |
 -- | ```purescript
--- | nub [1, 1, 2, 3, 3] = [1, 2, 3]
+-- | nub [1, 2, 1, 3, 3] = [1, 2, 3]
 -- | ```
 -- |
 nub :: forall a. Eq a => Array a -> Array a
@@ -872,7 +872,7 @@ nub = nubBy eq
 -- | by the specified equivalence relation, creating a new array.
 -- |
 -- | ```purescript
--- | nubBy (\a b → a `mod` 3 == b `mod` 3) [1, 3, 4, 5, 6] = [1,3,5]
+-- | nubBy (\a b -> a `mod` 3 == b `mod` 3) [1, 3, 4, 5, 6] = [1,3,5]
 -- | ```
 -- |
 nubBy :: forall a. (a -> a -> Boolean) -> Array a -> Array a
@@ -989,7 +989,7 @@ foreign import zipWith
 -- | `Applicative` functor.
 -- |
 -- | ```purescript
--- | sndChars = zipWithA (\a b → charAt 2 (a <> b))
+-- | sndChars = zipWithA (\a b -> charAt 2 (a <> b))
 -- | sndChars ["a", "b"] ["A", "B"] = Nothing -- since "aA" does not have a 3rd char
 -- | sndChars ["aa", "b"] ["AA", "BBB"] = Just ['A', 'B']
 -- | ```
@@ -1037,7 +1037,7 @@ unzip xs =
 -- | Perform a fold using a monadic step function.
 -- |
 -- | ```purescript
--- | foldM (\x y → Just (x + y)) 0 [1, 4] = Just 5
+-- | foldM (\x y -> Just (x + y)) 0 [1, 4] = Just 5
 -- | ```
 -- |
 foldM :: forall m a b. Monad m => (a -> b -> m a) -> a -> Array b -> m a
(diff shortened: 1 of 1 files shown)


src/Data/Array.purs
```diff
@@ -228,25 +272,47 @@ insertBy cmp x ys =
...
+-- |
 -- | Running time: `O(n)` where `n` is the length of the array
 tail :: forall a. Array a -> Maybe (Array a)
 tail = uncons' (const Nothing) (\_ xs -> Just xs)
 
 -- | Get all but the last element of an array, creating a new array, or
 -- | `Nothing` if the array is empty.
 -- |
+-- | ```purescript
+-- | init [1, 2, 3, 4] = Just [2, 3, 4]
```
fehrenbach: I think you mean [1, 2, 3] :)

src/Data/Array.purs
```diff
@@ -571,6 +842,11 @@ group' = group <<< sort
 
 -- | Group equal, consecutive elements of an array into arrays, using the
 -- | specified equivalence relation to detemine equality.
+-- |
+-- | ```purescript
+-- | groupBy (\a b → odd a && odd b) [1, 1, 2, 2, 3, 3] = [[1, 1], 2, 2, [3, 3]]
```
fehrenbach: For me this prints `[(NonEmpty 1 [1]),(NonEmpty 2 []),(NonEmpty 2 []),(NonEmpty 3 [3])]` which is not too pretty, but the 2s should be in singleton arrays at least: `[[1, 1], [2], [2], [3, 3]]`

csicar: I used the array-notation, because the example for `group` also used that notation

fehrenbach: Ah, I see. @garyb looks like you made the switch to `NonEmpty` (in 738f1b4d35eac922f21e64ef01e9f5002df8e92d). Did you intentionally not change the examples for `group` and `group'`?

garyb: No, it wasn't intentional :)

There is something nice about not showing the `NonEmpty` stuff in the example, as it does make it a bit clearer to see what is going on, but it's inaccurate obviously so I'm not sure which is better really. I guess we do show the exact output usually, so it probably should be in there. I don't know.

src/Data/Array.purs
```diff
@@ -434,29 +602,59 @@ filterA p =
...
+-- |
 catMaybes :: forall a. Array (Maybe a) -> Array a
 catMaybes = mapMaybe id
 
 -- | Apply a function to each element in an array, supplying a generated
 -- | zero-based index integer along with the element, creating an array
 -- | with the new elements.
+-- |
+-- | ```purescript
+-- | mapWithIndex (\index element -> (show index) <> element) ["Hello", "World"] = ["0Hello", "1World"]
```
hdgarrood: You have some redundant parens here

src/Data/Array.purs
```diff
@@ -708,6 +1053,12 @@ foldRecM f a array = tailRecM2 go a 0
         pure (Loop { a: res', b: i + 1 })
 
 -- | Find the element of an array at the specified index.
+-- |
+-- | ```purescript
+-- | unsafePartial $ unsafeIndex ["a", "b", "c"] 1 = "b"
+-- | unsafeIndex ["a", "b", "c"] 10 -- runtime error
```
hdgarrood: If I remember correctly, this won't produce a runtime error. Instead, the result will be `undefined` - of course, this is likely to cause a runtime error when the result is used, but it could also just cause the program to behave incorrectly but continue running. Would you mind checking this and fixing the example if necessary?

hdgarrood: Also, you'll need an `unsafePartial` here too, presumably?

csicar:
the problem with `group` etc. could be solved like this:
```purescript
group [1,1,2,2,1] == [(NonEmpty 1 [1]),(NonEmpty 2 [2]),(NonEmpty 1 [])]
map (fromFoldable) $ group [1,1,2,2,1] == [[1, 1], [2, 2], [1]]
```
that version would give the induition of the array-notation, while also being correct

hdgarrood:
I think that's a little verbose. I'd rather direct people to the NonEmpty documentation if they don't understand what `NonEmpty` means. (Of course, Pursuit already does this, because the NonEmpty type in the docs is linked.)

hdgarrood:
Oh also, the parens in eg `[(NonEmpty 2 [2])]` are redundant; the reason they appear in the REPL currently is just a shortcoming with the `Show` class (which will be fixed eventually). I think it would be better not to include them in these examples.

csicar:
I also tried to find examples for `some` and `many`, but could not understand them. Any ideas for simple example for those functions?

hdgarrood:
This PR is getting quite large, so I'd like to avoid it getting any larger. You can always send more later :) I'll have a think about `some` and `many` and get back to you. For now I'd like to just look over this one more time and then merge, as I think we've addressed everything raised now.

csicar:
sound good

src/Data/Array.purs
```diff
@@ -324,6 +429,12 @@ foreign import findIndexImpl
   -> Maybe Int
 
 -- | Find the last index for which a predicate holds.
+-- |
+-- | ```purescript
+-- | findLastIndex (_ > 0) [-10, -4, 2, 3] = Just 3
```
hdgarrood: I think it might be slightly better to use a type other than `Int` for the elements of the array here, actually, just to avoid potential confusion from people thinking `findLastIndex` might return an element of the array vs an index into it.

src/Data/Array.purs
```diff
@@ -556,21 +827,26 @@ span p arr =
...
+-- | group' [1,1,2,2,1] == [NonEmpty 1 [1,1],NonEmpty 2 [2]]
 -- | ```
 group' :: forall a. Ord a => Array a -> Array (NonEmpty Array a)
 group' = group <<< sort
 
 -- | Group equal, consecutive elements of an array into arrays, using the
 -- | specified equivalence relation to detemine equality.
+-- |
+-- | ```purescript
+-- | groupBy (\a b -> odd a && odd b) [1, 3, 2, 4, 3, 3] = [NonEmpty 1 [3], NonEmpty 2 [] , NonEmpty 4 [], NonEmpty 3 [3]]
```
hdgarrood: Can you please break this up over multiple lines? I don't think this will look great rendered as HTML. (You can check with `pulp docs -- --format html`)

src/Data/Array.purs
```diff
@@ -708,6 +1053,12 @@ foldRecM f a array = tailRecM2 go a 0
         pure (Loop { a: res', b: i + 1 })
 
 -- | Find the element of an array at the specified index.
+-- |
+-- | ```purescript
+-- | unsafePartial $ unsafeIndex ["a", "b", "c"] 1 = "b"
+-- | unsafePartial $ unsafeIndex ["a", "b", "c"] 10 -- runtime error
```
hdgarrood: I think this comment is still not quite correct.

csicar: I checked again in `psci`.
Result of evaluating:
```purescript
> unsafePartial $ unsafeIndex ["a", "b", "c"] 1
"b"

> unsafePartial $ unsafeIndex ["a", "b", "c"] 10
/home/.../.psci_modules/node_modules/Data.Show/foreign.js:30
  var l = s.length;
            ^

TypeError: Cannot read property 'length' of undefined
    at exports.showStringImpl (/home/.../.psci_modules/node_modules/Data.Show/foreign.js:30:13)
    at /home/.../node_modules/Control.Monad.Eff.Console/index.js:19:53
    at Object.<anonymous> (/home.../.psci_modules/node_modules/$PSCI/index.js:19:84)
    at Module._compile (module.js:641:30)
    at Object.Module._extensions..js (module.js:652:10)
    at Module.load (module.js:560:32)
    at tryModuleLoad (module.js:503:12)
    at Function.Module._load (module.js:495:3)
    at Module.require (module.js:585:17)
    at require (internal/module.js:11:18)
```

csicar: this is with version:
```bash
$ pulp --version                                                               
Pulp version 12.0.1
purs version 0.11.7 using /usr/local/bin/purs
```

hdgarrood: Right, but the error message indicates that the error comes from attempting to `Show` it - we might not necessarily do that in other situations.

hdgarrood: I'll try to come up with a good example

csicar: Oh yes, I see. Did not think of that..

hdgarrood: For example:
```purescript
module Main where

import Prelude
import Data.Array as Array
import Partial.Unsafe (unsafePartial)
import Control.Monad.Eff (Eff)
import Control.Monad.Eff.Console (CONSOLE, log)

main :: forall e. Eff (console :: CONSOLE | e) Unit
main = do
  let oops = unsafePartial (Array.unsafeIndex [true] 1)
  log $ if oops
    then "it was true"
    else "it was false"
```

Runs successfully and prints "it was false", because `undefined` is falsy.

src/Data/Array.purs
```diff
@@ -560,7 +566,12 @@ foreign import concat :: forall a. Array (Array a) -> Array a
 -- | into a single, new array.
 -- |
 -- | ```purescript
--- | concatMap (split $ Pattern " ") ["Hello World", "other thing"] = ["Hello", "World", "other", "thing"]
+-- | concatMap (split $ Pattern " ") ["Hello World", "other thing"]
+-- | ```
```
hdgarrood: Sorry to nitpick but I think I prefer the style you used in e.g. `alterAt`, where it's still all contained in a single code block.

src/Data/Array.purs
hdgarrood: Since this is a little subtle, maybe it would be best to remove the out-of-range example, and instead explain what happens if you try to `unsafeIndex` with an out-of-range index in prose. How about:

> Using `unsafeIndex` with an out-of-range index will not immediately raise a runtime error. Instead, the result will be `undefined`. Most attempts to subsequently use the result will cause a runtime error, of course, but this is not guaranteed, and is dependent on the backend; some programs will continue to run as if nothing is wrong. For example, in the JavaScript backend, the expression `unsafePartial (unsafeIndex [true] 1)` has type `Boolean`; since this expression evaluates to `undefined`, attempting to use it in an `if` statement will cause the `else` branch to be taken.

hdgarrood:
Fantastic, thank you very much for this!

csicar:
thank's for merging!

Issue #118, opened 2017-11-30
Add unsafeThaw to Data.Array.ST?


hdgarrood:
I would like to be able to create an array using one of the existing functions for immutable arrays (e.g. `range`), and then use it locally within a function, but not return a reference to it. I would also like to be able to do this with a minimum of copying. Currently I think the best I can do is to use `thaw` or `withArray`, which make a copy of it; if I know that no references to my array can escape I should be able to get away without copying it. Could we possibly add an `unsafeThaw`, which has the same type as `thaw` but just returns a reference to the same array instead of copying?

paf31:
👍 Sounds good to me.

matthewleon:
Going to make a PR for this since nobody has done it, but I think it would also be nice to have PRs to implement the immutable array functions for ST arrays, even if internally they just use `unsafeThaw`. Any thoughts on this approach? I could make a separate issue for it.

hdgarrood:
Resolved by #123.

PR #123, opened 2018-01-19
ST.unsafeThaw

matthewleon:
addresses https://github.com/purescript/purescript-arrays/issues/118

### DIFF
--- a/src/Data/Array/ST.purs
+++ b/src/Data/Array/ST.purs
@@ -14,8 +14,10 @@ module Data.Array.ST
   , modifySTArray
   , pushAllSTArray
   , spliceSTArray
-  , freeze, thaw
+  , freeze
+  , thaw
   , unsafeFreeze
+  , unsafeThaw
   , toAssocArray
   ) where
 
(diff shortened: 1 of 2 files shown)


src/Data/Array/ST.purs
```diff
@@ -66,6 +68,11 @@ withArray f xs = do
 unsafeFreeze :: forall a r h. STArray h a -> Eff (st :: ST h | r) (Array a)
 unsafeFreeze = pure <<< (unsafeCoerce :: STArray h a -> Array a)
 
+-- | O(1) Convert an immutable array to a mutable array, without copying. The mutable
+-- | array must not be mutated.
```
hdgarrood: I think writing "the mutable array must not be mutated" seems to defeat the point of this function: if you have an `Array a` and you want to convert it into an `STArray h a`, that's probably because you want to mutate it, right? I think it would make more sense to say "the input array must not be used afterwards".

matthewleon: Right. Will fix.

matthewleon:
Modified according to review comment.

hdgarrood:
Thanks!

PR #124, opened 2018-01-19
reuse bindFlipped for concatMap

matthewleon:
Admittedly, this is very nitpicky, but it makes sense to me to reuse a
function defined elsewhere.


hdgarrood:
I'd rather not, since the way I see it, `bindFlipped` is only really there so that it can be used infix as `=<<`; I've never seen `bindFlipped` written out like this elsewhere so I think the way it is now is easier to understand.

matthewleon:
fair enough

Issue #122, opened 2018-01-17
The absence of `map` function


ivan-kleshnin:
No doubt it's intentional. Same stuff for `purescript-lists`. 
My guess is that `map` is not exposed to not collide with `Prelude.map`. This has downsides, though.

1. Educational. I can't say "import `Data.Array as Array` and use concrete types before you're ready to functors", because the most important function isn't there.

2. Namespacing. Special rule for `map` in `Array.concatMap _ >>> map _ >>> Array.whatever _` kind of constructions is annoying. You can say it's a matter of habit, but in JS I use qualified imports almost exclusively. So I expected to do the same in PureScript. Maybe it's not a pragmatic approach (because operators) – you tell me.

*Btw. can we namespace operators like `Array.(..)` or something?*

3. Consistency. The presence of generic `map` makes `Array.map` unnecessary? Well, there's `Control.Bind.bindFlipped` and there's `Array.concatMap` so no.

matthewleon:
Would you resolve this through simply exposing a `map = Functor.map` with a monomorphic type signature?

As for namespacing, is this something you can't accomplish with `import Data.Array as Array`?

ivan-kleshnin:
> Would you resolve this through simply exposing a map = Functor.map with a monomorphic type signature?

I would copy `filter`: https://github.com/purescript/purescript-arrays/blob/master/src/Data/Array.purs#L576-L583

```
foreign import filter :: forall a. (a -> Boolean) -> Array a -> Array a
->
foreign import map :: forall a. (a -> b) -> Array a -> Array b
``` 

> As for namespacing, is this something you can't accomplish with import Data.Array as Array?

Current `Array` doesn't export `map`.

matthewleon:
> I would copy filter: https://github.com/purescript/purescript-arrays/blob/master/src/Data/Array.purs#L576-L583

The difference here is that `map` is already defined for Arrays in the Prelude: https://github.com/purescript/purescript-prelude/blob/v3.1.0/src/Data/Functor.purs#L42

So doing the `foreign import` approach would mean having two copies of the same code. This is why I suggest the re-export based approach. Is there a disadvantage to that?

matthewleon:
@ivan-kleshnin please let me know if https://github.com/purescript/purescript-arrays/pull/125 does what you want

hdgarrood:
[...]

PR #125, opened 2018-01-19
re-export Functor.map

matthewleon:
addresses https://github.com/purescript/purescript-arrays/issues/122


hdgarrood:
I’m :-1: for this - `map` is already in Prelude, which is likely to be imported already anyway, so I think this doesn’t add much; however it does enable nonsense like `Array.map (_+1) (_+2) 0 == 3`.

matthewleon:
Closing.

Issue #129, opened 2018-01-23
deprecate verbose ST function names


matthewleon:
`pushSTArray`, `pokeSTArray`, etc. could be exported as `push` and `poke`. Library users could use qualified imports for disambiguation.

hdgarrood:
Big :+1: from me, this has vaguely bothered me for a while. I’d like to wait and see what others think though.

garyb:
No argument from me, I always import this stuff qualified anyway.

garyb:
Resolved in the 0.12 branch

PR #130, opened 2018-01-23
STArray: in-place sort

matthewleon:


matthewleon:
Using a simpler naming convention here, as proposed in https://github.com/purescript/purescript-arrays/issues/129

I'll change these to the more verbose form if people prefer.

matthewleon:
@hdgarrood any chance of a release with this and `unsafeThaw`?

PR #131, opened 2018-01-23
use purescript-functions to simplify sort FFI

matthewleon:


hdgarrood:
Does this improve performance? If not I’m tempted to leave it as is.

matthewleon:
No perf change from this, was just a thought that came up when I made the new `ST` sort functions. Closing.

PR #146, opened 2018-08-16
Fix #145: Incompatibility with `purs bundle`

### LINKED ISSUE #145: Incompatibility with `purs bundle`
The following line:

https://github.com/purescript/purescript-arrays/blob/d81e97137d2040f2fb42bf66124eb1d96e5dc1a7/src/Data/Array.js#L43

causes runtime errors when compiled with `purs bundle`, because `replicatePolyfill` gets eliminated as dead code.

I'm not sure if it is a `purs bundle` bug, since I don't understand exactly the spec of the code elimination it performs (in particular, what shapes of exports it supports).

maximedenes:
This fixes code generated by `purs bundle` for navigators that don't provide `Array.prototype.fill`, like IE11.

### DIFF
--- a/src/Data/Array.js
+++ b/src/Data/Array.js
@@ -18,7 +18,7 @@ exports.range = function (start) {
   };
 };
 
-var replicate = function (count) {
+var replicateFill = function (count) {
   return function (value) {
     if (count < 1) {
       return [];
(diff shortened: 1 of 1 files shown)


garyb:
Since a reproducible example was provided in #145 I can see that it works now - so thanks!

PR #150, opened 2018-10-05
Align the `nubByEq` example with the `unionBy` example and the `deleteBy` example

ewaldgrusk:
Sorry, I can’t unsee it. 😅

Please feel free to close this pull request if you think that this change is unnecessary.


ewaldgrusk:
By the way, most of the examples in the documentation use a single `=` to denote that the expression on the left-hand side evaluates to the expression on the right-hand side. However, some examples (e.g. `span`, `group`, and `nubBy`) use `==` instead.

And now I’m wondering if this is intentional. 😅

PR #152, opened 2018-12-14
Add runSTArray for efficiently constructing arrays using ST.

Dretch:
Haskell has this already. Also add a test for withArray.

I think this makes sense - please reject this pull request if not!

### DIFF
--- a/src/Data/Array/ST.purs
+++ b/src/Data/Array/ST.purs
@@ -5,6 +5,7 @@
 module Data.Array.ST
   ( STArray(..)
   , Assoc
+  , runSTArray
   , withArray
   , empty
   , peek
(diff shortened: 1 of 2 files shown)


hdgarrood:
As it happens, we previously had this (and not `unsafeFreeze`), but then we removed it at the same time we added `unsafeFreeze`; of course you can reconstruct this function in terms of `unsafeFreeze` (as you have here). I didn't realise at the time that this function is useful to have in addition to `unsafeFreeze` though, since of course this function allows you to obtain the benefits of `unsafeFreeze` while retaining more type safety. See #87, #89 

So :+1: from me, although I would suggest calling it just `run`, as we've tried to avoid module name / function name duplication in this module.

Dretch:
@hdgarrood it is interesting to know the history behind this, thanks.

I have renamed `runSTArray` to `run`, as requested.

PR #156, opened 2019-03-31
Using head and isNothing to compute whether a given array is null or not

aravindgopall:
in O(1).


garyb:
I'm pretty sure `length` will be faster than this, it's also probably O(1), since it's using the JS `length` function.

natefaubion:
Yes, `length` is `O(1)`, and it also doesn't allocate a `Maybe` box.

PR #157, opened 2019-04-02
added: pop, shift, unshift, unshiftAll to Data.Array.ST

### LINKED ISSUE #151: add pop, shift, unshift to STArray
Just like the methods of the same names on a JS Array:

- `pop` removes the last item of an array and returns that item
- `shift` removes the first item of an array and returns that item
- `unshift` takes an item and sets it as the first element of the array, shifting everything else along by 1

8084:
closes #151

### DIFF
--- a/src/Data/Array/ST.js
+++ b/src/Data/Array/ST.js
@@ -28,10 +28,38 @@ exports.poke = function (i) {
   };
 };
 
+exports.popImpl = function (just) {
+  return function (nothing) {
+    return function (xs) {
+      return function () {
+        return xs.length ? just(xs.pop()) : nothing;
+      };
+    };
+  };
+};
+
 exports.pushAll = function (as) {
   return function (xs) {
     return function () {
       return xs.push.apply(xs, as);
+    };
+  };
+};
+
+exports.shiftImpl = function (just) {
+  return function (nothing) {
+    return function (xs) {
+      return function () {
+        return xs.length ? just(xs.shift()) : nothing;
+      };
+    };
+  };
+};
+
+exports.unshiftAll = function (as) {
+  return function (xs) {
+    return function () {
+      return xs.unshift.apply(xs, as);
     };
   };
 };
--- a/src/Data/Array/ST.js
+++ b/src/Data/Array/ST.js
@@ -32,7 +32,7 @@ exports.popImpl = function (just) {
   return function (nothing) {
     return function (xs) {
       return function () {
-        return xs.length ? just(xs.pop()) : nothing;
+        return xs.length > 0 ? just(xs.pop()) : nothing;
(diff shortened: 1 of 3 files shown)


src/Data/Array/ST.purs
```diff
@@ -83,6 +87,12 @@ thaw = copyImpl
 sort :: forall a h. Ord a => STArray h a -> ST h (STArray h a)
 sort = sortBy compare
 
+-- | Remove the first element from an array and return that element.
+shift :: forall h a. STArray h a -> ST h (Maybe a)
+shift = shiftImpl Just Nothing
+
+foreign import shiftImpl :: forall h a. (a -> Maybe a) -> Maybe a -> STArray h a -> ST h (Maybe a)
```
hdgarrood: Can we use rank-n types to make this a little safer please? If you write e.g. `(forall b. b -> Maybe b)` for the `Just` argument, then there are only two possibilities for the function which can go in there (i.e. either `const Nothing` or `Just`). This also makes it a little clearer how this is intended to be used. Also could you please put each argument on its own line to match the formatting for existing FFI imports in this file?

src/Data/Array/ST.js
```diff
@@ -28,6 +28,16 @@ exports.poke = function (i) {
   };
 };
 
+exports.popImpl = function (just) {
+  return function (nothing) {
+    return function (xs) {
+      return function () {
+        return xs.length ? just(xs.pop()) : nothing;
```
hdgarrood: I'm not particularly keen on relying on the falsiness of 0. Can you use `xs.length > 0` for the condition here instead please?

src/Data/Array/ST.js
```diff
@@ -36,6 +46,24 @@ exports.pushAll = function (as) {
   };
 };
 
+exports.shiftImpl = function (just) {
+  return function (nothing) {
+    return function (xs) {
+      return function () {
+        return xs.length ? just(xs.shift()) : nothing;
```
hdgarrood: Same here

src/Data/Array/ST.purs
```diff
@@ -136,6 +146,12 @@ foreign import peekImpl
 -- | Change the value at the specified index in a mutable array.
 foreign import poke :: forall h a. Int -> a -> STArray h a -> ST h Boolean
 
+-- | Remove the last element from an array and return that element.
+pop :: forall h a. STArray h a -> ST h (Maybe a)
+pop = popImpl Just Nothing
+
+foreign import popImpl :: forall h a. (a -> Maybe a) -> Maybe a -> STArray h a -> ST h (Maybe a)
```
hdgarrood: Same here re rank-n types and formatting

Issue #155, opened 2019-02-19
`sort`/`sortBy` Does Not Use The Correct Ordering & Violates Parametricity


eric-corumdigital:
> All undefined elements are sorted to the end of the array, with no call to compareFunction.

— https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Description

Therefore, sorting on any type inhabited by `undefined` will not use the correct ordering (except by coincidence). This is also a parametricity violation.

The only truthful use of native sort is on JavaScript values.

I propose we replace the implementation with a correct one. Yes, it will be slower. We might define `unsafeSort` to use the native sort with the same typing as `sort`. Then we may also have a different sort definition typed on JavaScript values (which is Foreign).

eric-corumdigital:
I bear gifts:

```
function mergeFromTo(compare, ordToNum, xs1, xs2, from, to) {

  var mid;
  var i;
  var j;
  var k;
  var x;
  var y;
  var c;

  mid = from + ((to - from) >> 1);
  if (mid - from > 1) mergeFromTo(compare, ordToNum, xs2, xs1, from, mid);
  if (to - mid > 1) mergeFromTo(compare, ordToNum, xs2, xs1, mid, to);
  
  i = from;
  j = mid;
  k = from;
  while (i < mid && j < to) {
    x = xs2[i];
    y = xs2[j];
    c = ordToNum(compare(x, y));
    if (c > 0) {
      xs1[k++] = y;
      ++j;
    }
    else if (c === 0) {
      xs1[k++] = x;
      xs1[k++] = y;
      ++i;
      ++j;
    }
    else {
      xs1[k++] = x;
      ++i;
    }
  }
  while (i < mid) {
    xs1[k++] = xs2[i++];
  }
  while (j < to) {
    xs1[k++] = xs2[j++];
  }
  
}

function mergeSort(compare, ordToNum, xs) {

  var out;

  if (xs.length < 2) return xs;
  
  out = xs.slice(0);
  mergeFromTo(compare, ordToNum, out, xs.slice(0), 0, xs.length);

  return out;
};
```

```
foreign import mergeSort ∷ ∀ a. Fn3 (Fn2 a a Ordering) (Ordering → Number) (Array a) (Array a)
```

Duplicates the input array twice. One copy is garbage and the other becomes the sorted array.

[...]

PR #158, opened 2019-06-28
Explicitly document sort stability

ford-prefect:
As not all users of this library might be aware of the affact that
Array.prototype.sort() can be stable or unstable depending on the actual
JavaScript runtime being used (and potentially the length of the array),
let's explicitly document this to avoid people from getting tripped up.

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -681,6 +681,9 @@ modifyAtIndices is f xs =
 
 -- | Sort the elements of an array in increasing order, creating a new array.
 -- |
+-- | The underlying sort may or may not be stable, depending on your JavaScript
+-- | runtime.
+-- |
 -- | ```purescript
 -- | sort [2, -3, 1] = [-3, 1, 2]
 -- | ```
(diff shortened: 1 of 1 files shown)


sharno:
I think this is unnecessary now given this? :

https://github.com/tc39/ecma262/pull/1340

hdgarrood:
Thanks for the link! I think we should document sort’s stability either way (whether it’s stable or not). #155 is likely going to be in the next release anyway though.

JordanMartinez:
Does #155 still need to be implemented? Or will merging this suffice due to the changes made to sort stability since this was made?

thomashoneyman:
With #195 it's unclear to me whether this is still relevant. Should this be closed?

ford-prefect:
If the new `sortByImpl` is a stable sort, then we can close this.

hdgarrood:
This should now be closed as the stability of sort no longer depends on your JS runtime. However, we should still document the stability of our sort, whether it is stable or not. I’m not sure if the new sort algorithm is stable - we should verify this, and if it is, we should document this and add tests for stability (so that we don’t accidentally lose stability later).

JordanMartinez:
I opened #196 to track the sort stability stuff.

PR #159, opened 2019-09-12
Replace use of unsafeCoerce in freeze/thaw functions with discrete fo…

andyarvanitis:
…reign functions

The assumption that you can `unsafeCoerce` from an immutable to a mutable array and back doesn't hold for the golang backend – and maybe won't for other backends either, particularly ones that actually have immutable array types. Please let me know if you'd like any more info, or if you'd like me to create an issue for the change in addition to this PR. Thanks.

Reference: https://github.com/andyarvanitis/purescript-native-go-ffi/pull/9

### DIFF
--- a/src/Data/Array/ST.js
+++ b/src/Data/Array/ST.js
@@ -76,11 +76,27 @@ exports.splice = function (i) {
   };
 };
 
-exports.copyImpl = function (xs) {
+exports.unsafeFreeze = function (xs) {
+  return function () {
+    return xs;
+  };
+};
+
+exports.unsafeThaw = function (xs) {
+  return function () {
+    return xs;
+  };
+};
+
+function copyImpl(xs) {
   return function () {
     return xs.slice();
   };
-};
+}
+
+exports.freeze = copyImpl;
+
+exports.thaw = copyImpl;
 
 exports.sortByImpl = function (comp) {
   return function (xs) {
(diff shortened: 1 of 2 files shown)


hdgarrood:
Hm, I'm not entirely sure about this: the docs specifically say that this is an O(1) operation, and I think there's a fair bit of code out there which assumes that `STArray` and `Array` have the same representation. In fact if they don't have the same representation there isn't really any reason for `unsafeFreeze` and `unsafeThaw` to exist at all.

sharno:
The operation in JS and Go will still be O(1) for unsafe conversions, the problem is that Go discriminates between passing a parameter to a function by value (shallow copying) vs by pointer. This makes the usage of `unsafeCoerce` for both `unsafeFreeze` and `unsafeThaw` hard to handle in Go.

This is because `unsafeThaw` needs to return the pointer to deal with `STArray` as a pointer everywhere, while `unsafeFreeze` will return the the `Array` by value (shallow copying the pointer everywhere it's passed to a function.

So `unsafeFreeze` and `unsafeThaw` will still have the advantage of being O(1) vs their safe counterparts which take O(n) to copy

hdgarrood:
Ah I see, in that case this ought to be fine. Just to check - when you say "shallow copying", does that mean that it is just the pointer address which is being copied, not the contents of the array?

sharno:
Yeah I meant that only the pointer address is copied not the contents

sharno:
@hdgarrood Is there any updates if this will be merged?

andyarvanitis:
Thanks, Harry!

hdgarrood:
Welcome :) would it help for me to cut a release?

andyarvanitis:
Sure, please, if you don't mind. I certainly don't have any dependencies on the current release.

hdgarrood:
:+1: I've just published v5.3.1

PR #171, opened 2020-05-11
Update README.md for spago

csicar:


milesfrain:
Do we even want to list the `bower` option anymore?

hdgarrood:
I agree, replacing `bower` with `spago` seems like the best approach to me.

thomashoneyman (approved):

PR #172, opened 2020-05-22
Running time note for snoc

milesfrain:
It would be nice to be upfront about the runtime benefit of `snoc`, rather than having users confirm this for themselves by looking in `Arrays.js`.

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -228,7 +228,7 @@ foreign import length :: forall a. Array a -> Int
 -- | cons 1 [2, 3, 4] = [1, 2, 3, 4]
 -- | ```
 -- |
--- | Note, the running time of this function is `O(n)`.
+-- | Note, the running time of this function is `O(n)`. For an `O(1)` alternative, see `snoc`.
 foreign import cons :: forall a. a -> Array a -> Array a
 
 -- | An infix alias for `cons`.
@@ -237,7 +237,7 @@ foreign import cons :: forall a. a -> Array a -> Array a
 -- | 1 : [2, 3, 4] = [1, 2, 3, 4]
 -- | ```
 -- |
--- | Note, the running time of this function is `O(n)`.
+-- | Note, the running time of this function is `O(n)`. For an `O(1)` alternative, see `snoc`.
 infixr 6 cons as :
 
 -- | Append an element to the end of an array, creating a new array.
@@ -246,6 +246,7 @@ infixr 6 cons as :
 -- | snoc [1, 2, 3] 4 = [1, 2, 3, 4]
 -- | ```
 -- |
+-- | The running time of this function is `O(1)`.
 foreign import snoc :: forall a. Array a -> a -> Array a
 
 -- | Insert an element into a sorted array.


garyb:
`snoc` is `O(n)` also - all the array-modifying functions are at least `O(n)` since they need to copy the whole array first.

hdgarrood:
> all the array-modifying functions are at least O(n) since they need to copy the whole array first.

Perhaps it would be best to expand and clarify the note at the top of the module instead; this is useful information and it's left implicit by that note right now.

milesfrain:
My mistake. I assumed `slice` returned a new reference to the same underlying array (rather than copying), although I now see how that would be undesirable in lots of situations.

Could an optimizer figure out that this is our intention, and avoid copying the array upon push?

hdgarrood:
I highly doubt it; I can’t see that being safe unless it’s handled by the JS VM.

garyb:
Yeah, I'm not sure why it's implemented with slice rather than it being `x.concat([y])`, I assume it was benchmarked and the push version was faster.

The only way we could safely avoid copying arrays is with linear types, so we know they're not reused, etc.

milesfrain:
> Perhaps it would be best to expand and clarify the note at the top of the module instead; this is useful information and it's left implicit by that note right now.

Applied that change in the latest commit.

Now what should the runtime notes be for the other functions? Should we either:
* Remove all additional `O(n)` notes
* Make sure every function has a runtime note

For the latter, which note format do you prefer? I see these three varieties in the docs:
* > Note, the running time of this function is `O(n)`.
* > Running time: `O(n)`.
* > Running time: `O(n)` where `n` is the length of the array

hdgarrood:
I would prefer the third. I think we should explicitly say what `n` is whenever we are using big O notation, ideally.

src/Data/Array.purs
```diff
@@ -4,7 +4,9 @@
 -- | `Data.Sequence` instead, which might give better performance for certain
 -- | use cases. This module is useful when integrating with JavaScript libraries
 -- | which use arrays, but immutable arrays are not a practical data structure
--- | for many use cases due to their poor asymptotics.
+-- | for many use cases due to their poor asymptotics. The running time of all
```
hdgarrood, line 7: Big O tells you the computational complexity, which is not quite the same as running time (although of course they are related). Additionally, "The running time" is singular so the verb should be "is" rather than "are" here. However, I think it might be easiest to sidestep these issues by writing "All array modification operations (...) are at least as slow as `O(n)` ...".

hdgarrood:
I think most functions should have complexity noted, although not all. For instance, it doesn't really make sense to give `singleton` one, because there's no sensible way of defining the size of the input. Also, we can't give one for `fromFoldable`, because it depends on the `Foldable` instance in use.

milesfrain:
> I think most functions should have complexity noted

Added some more running time notes. Wasn't sure about some of these though (`updateAtIndices`, `sort`, etc.), so just left those blank.

Also seems like `filterA` could be O(n^2), depending on the predicate, so just noted "At least O(n)".

PR #179, opened 2020-09-27
Add `splitAt`

maxdeviant:
This PR adds a `splitAt` function for `Array`s and `NonEmptyArray`s.

### Prior Art

**Haskell:** [`splitAt`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Prelude.html#v:splitAt)
**fp-ts:** [`splitAt`](https://gcanti.github.io/fp-ts/modules/Array.ts.html#splitat)

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -601,15 +601,25 @@ foreign import partition
   -> Array a
   -> { yes :: Array a, no :: Array a }
 
--- | Splits an array into two pieces, where the first array has `n` elements
--- | and the second array has the remaining elements.
+-- | Splits an array into two subarrays, where `before` contains the elements
+-- | up to (but not including) the given index, and `after` contains the rest
+-- | of the elements, from that index on.
 -- |
 -- | ```purescript
--- | splitAt 3 [1, 2, 3, 4, 5] == Tuple [1, 2, 3] [4, 5]
+-- | >>> splitAt 3 [1, 2, 3, 4, 5]
+-- | { before: [1, 2, 3], after: [4, 5] }
 -- | ```
-splitAt :: forall a. Int -> Array a -> Tuple (Array a) (Array a)
-splitAt n xs | n <= 0 = Tuple [] xs
-splitAt n xs = Tuple (slice 0 n xs) (slice n (length xs) xs)
+-- |
+-- | Thus, the length of `(splitAt i arr).before` will equal either `i` or
+-- | `length arr`, if that is shorter. (Or if `i` is negative the length will
+-- | be 0.)
+-- |
+-- | ```purescript
+-- | splitAt 3 [1, 2, 3, 4, 5] == { before: [1, 2, 3], after: [4, 5] }
+-- | ```
+splitAt :: forall a. Int -> Array a -> { before :: Array a, after :: Array a }
+splitAt i xs | i <= 0 = { before: [], after: xs }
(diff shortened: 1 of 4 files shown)


test/Test/Data/Array.purs
```diff
@@ -202,6 +202,14 @@ testArray = do
...
   assert $ A.filter odd (A.range 0 10) == [1, 3, 5, 7, 9]
 
+  log "splitAt should split the array at the given number of elements"
+  assert $ A.splitAt 3 [1, 2, 3, 4, 5] == Tuple [1, 2, 3] [4, 5]
+  assert $ A.splitAt 1 [1, 2, 3] == Tuple [1] [2, 3]
+  assert $ A.splitAt 3 [1, 2, 3] == Tuple [1, 2, 3] []
+  assert $ A.splitAt 4 [1, 2, 3] == Tuple [1, 2, 3] []
+  assert $ A.splitAt 0 [1, 2, 3] == Tuple [] [1, 2, 3]
+  assert $ A.splitAt (-1) [1, 2, 3] == Tuple [] [1, 2, 3]
+
```
JordanMartinez, line 212: There should be a test here for when the array argument is empty. I assume `splitAt n []` produces `Tuple [] []`.

src/Data/Array.purs
```diff
@@ -600,6 +601,16 @@ foreign import partition
   -> Array a
   -> { yes :: Array a, no :: Array a }
 
+-- | Splits an array into two pieces, where the first array has `n` elements
+-- | and the second array has the remaining elements.
+-- |
+-- | ```purescript
+-- | splitAt 3 [1, 2, 3, 4, 5] == Tuple [1, 2, 3] [4, 5]
+-- | ```
```
JordanMartinez, line 609: Docs should specify what happens when an empty array is the argument

hdgarrood:
I wonder if we should have this return a record `{ before :: Array a, after :: Array a }` rather than a `Tuple`, since we're already doing something similar with `partition`, and the `splitAt` for strings already returns a record: https://pursuit.purescript.org/packages/purescript-strings/4.0.1/docs/Data.String.CodePoints#v:splitAt

JordanMartinez:
I'd agree with that. It also removes the need to document which array is the before/after part (though that itself is likely obvious).

test/Test/Data/Array.purs
maxdeviant: Added in 6f414d6.

src/Data/Array.purs
maxdeviant: Added in 6f414d6.

maxdeviant:
`splitAt` now returns `{ before :: Array a, after :: Array a }` instead of a `Tuple`.

Funnily enough, I had initially considered doing this, but was hung up on the `before` terminology since I was thinking of it in terms of `take n elements`, which is inclusive.

It wasn't until just now where I realized that, conceptually, taking `n` elements from an array (inclusive) is the same as taking elements up to that index (exclusive).

JordanMartinez (approved):

JordanMartinez:
> It wasn't until just now where I realized that, conceptually, taking n elements from an array (inclusive) is the same as taking elements up to that index (exclusive).

I see it more like this:
```purescript
let { before, after} = splitAt i array
in 
  (before == take i array) && (after == drop i array)
```

hdgarrood (approved):

PR #180, opened 2020-10-06
replace FFI cons, snoc, drop, take with purescript implementations

sharno:
I feel it's a good idea to replace FFI implementations with equivalent Purescript ones, mainly to make it easier for other backends to use the library with less FFI code and for the type safety. What do you think?

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -122,9 +122,9 @@ import Control.Alternative (class Alternative)
 import Control.Lazy (class Lazy, defer)
 import Control.Monad.Rec.Class (class MonadRec, Step(..), tailRecM2)
 import Control.Monad.ST as ST
+import Data.Array.NonEmpty.Internal (NonEmptyArray)
 import Data.Array.ST as STA
 import Data.Array.ST.Iterator as STAI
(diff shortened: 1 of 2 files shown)


src/Data/Array.purs
```diff
@@ -229,7 +229,8 @@ foreign import length :: forall a. Array a -> Int
 -- | ```
 -- |
 -- | Note, the running time of this function is `O(n)`.
-foreign import cons :: forall a. a -> Array a -> Array a
+cons :: forall a. a -> Array a -> Array a
+cons x xs = ST.run (STA.withArray (STA.splice 0 0 [x]) xs)
```
hdgarrood, line 233: I think `[x] <> xs` would be equally good as the implementation here, and same for snoc.

hdgarrood (commented):
Completely agree with reducing the amount of FFI here, yes.

src/Data/Array.purs
sharno: I think that's a good idea for `cons`, but I tried to benchmark `.concat` which `<>` uses and `.slice().push()` that we use for `snoc` and `.slice().push()` seems faster around 75% in chrome

hdgarrood (approved):

PR #181, opened 2020-10-20
Update to v0.14.0-rc3

paluh:
Backlinking to purescript/purescript#3942

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,8 @@ node_js: stable
 env:
   - PATH=$HOME/purescript:$PATH
 install:
-  - TAG=$(basename $(curl --location --silent --output /dev/null -w %{url_effective} https://github.com/purescript/purescript/releases/latest))
+  # - TAG=$(basename $(curl --location --silent --output /dev/null -w %{url_effective} https://github.com/purescript/purescript/releases/latest))
+  - TAG=v0.14.0-rc3
   - curl --location --output $HOME/purescript.tar.gz https://github.com/purescript/purescript/releases/download/$TAG/linux64.tar.gz
   - tar -xvf $HOME/purescript.tar.gz -C $HOME/
   - chmod a+x $HOME/purescript
--- a/bower.json
+++ b/bower.json
@@ -15,22 +15,22 @@
     "package.json"
   ],
   "dependencies": {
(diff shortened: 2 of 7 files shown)


paluh:
Includes #176

srghma:
related https://github.com/purescript/purescript-arrays/pull/183

JordanMartinez:
@paluh please merge #183 into this PR

kl0tl:
We could replace some uses of unsafeCoerce by Safe.Coerce.coerce in the implementations of Data.Array.NonEmpty.unsafeFromArray, Data.Array.NonEmpty.toArray and Data.Array.groupBy. This requires the NonEmptyArray constructor to be in scope so we should be careful not to re-export it from Data.Array.NonEmpty.

Sadly we can’t remove the last unsafeCoerce from Data.Array.unsafeFromArrayF implementation without strengthening its type with a `Coercible (f (Array a)) (f (NonEmptyArray a))` constraint, so we should probably refrain from doing it (at least until we have constraint kinds and quantified constraints so we can define `type Representational f = forall a b. Coercible a b => Coercible (f a) (f b)`).

paluh:
@JordanMartinez I've included #183 . Do we want to include suggestions by @kl0tl here too?

JordanMartinez:
I think we should do @kl0tl's suggestions in a separate PR. They aren't strictly needed to update this library to v0.14.0, but are welcome improvements.

JordanMartinez:
@paluh All dependencies are now satisfied again. We're still getting an error due to the [following line](https://travis-ci.org/github/purescript/purescript-arrays/builds/739085587#L741):
```
[1/1 WarningParsingModule] src/Data/Array/ST.purs:34:30

  34  import Control.Monad.ST (ST, kind Region)

                                   ^^^^^^^^^^^

  

  Kind imports are deprecated and will be removed in a future release. Omit the 'kind' keyword instead.

```
We just need to delete the `kind` word.

JordanMartinez (approved):
CI now builds. I've also looked through the changes made and have verified that no unicode syntax is being used.

We'll need a second approval before we can merge. Once merged, this will unblock `purescript-transformers`.

thomashoneyman (approved):

PR #182, opened 2020-10-24
Add unsafe versions of delete, insert, update, and modify

### LINKED ISSUE #167: Expose unsafe versions of `insertAt`, `deleteAt`, and `modifyAt`
Libraries that wish to build upon this one have to implement their own FFI when trying to maximize performance. So, could we expose the FFI here, so that it's more backend-friendly?

JordanMartinez:
Fixes #167 

Tests haven't been implemented yet since I wanted to ensure this would be accepted and the approach is correct. I wasn't sure whether I should implement this via `Data.Array.ST` or not.

### DIFF
--- a/src/Data/Array.js
+++ b/src/Data/Array.js
@@ -148,7 +148,7 @@ exports._insertAt = function (just) {
   };
 };
 
-exports._unsafeInsertAt = function (i) {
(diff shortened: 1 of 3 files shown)


JordanMartinez:
I still think this should be implemented, but I'm closing for now.

PR #183, opened 2020-10-24
fix: Fix `Foldable1 NonEmptyArray` instance (requires https://github.com/purescript/purescript-nonempty/pull/39)

srghma:
…

### DIFF
--- a/src/Data/Array/NonEmpty/Internal.js
+++ b/src/Data/Array/NonEmpty/Internal.js
@@ -1,11 +1,21 @@
 "use strict";
 
-exports.fold1Impl = function (f) {
+exports.foldr1Impl = function (f) {
   return function (xs) {
     var acc = xs[0];
     var len = xs.length;
     for (var i = 1; i < len; i++) {
       acc = f(acc)(xs[i]);
+    }
+    return acc;
+  };
+};
+
+exports.foldl1Impl = function (f) {
+  return function (xs) {
+    var acc = xs[xs.length - 1];
+    for (var i = xs.length - 2; i >= 0; i--) {
+      acc = f(xs[i])(acc);
     }
     return acc;
   };
--- a/src/Data/Array/NonEmpty/Internal.js
+++ b/src/Data/Array/NonEmpty/Internal.js
@@ -2,10 +2,9 @@
 
 exports.foldr1Impl = function (f) {
   return function (xs) {
-    var acc = xs[0];
-    var len = xs.length;
-    for (var i = 1; i < len; i++) {
-      acc = f(acc)(xs[i]);
+    var acc = xs[xs.length - 1];
+    for (var i = xs.length - 2; i >= 0; i--) {
+      acc = f(xs[i])(acc);
     }
     return acc;
   };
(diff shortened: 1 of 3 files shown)


src/Data/Array/NonEmpty/Internal.js
```diff
@@ -1,6 +1,6 @@
 "use strict";
 
-exports.fold1Impl = function (f) {
+exports.foldr1Impl = function (f) {
```
JordanMartinez, line 3: Isn't this the implementation for `foldl1`?

test/Test/Data/Array/NonEmpty.purs
```diff
@@ -297,12 +297,22 @@ testNonEmptyArray = do
...
 
-  log "foldMap1 should work"
-  assert $ foldMap1 Additive (fromArray [1, 2, 3, 4]) == Additive 10
+  log "foldr1 should work"
+  assert $ foldr1 (\l r -> "(" <> l <> r <> ")") (fromArray ["a", "b", "c", "d"]) == "(((ab)c)d)"
+  assert $ foldr1 (\l r -> "(" <> l <> r <> ")") (fromArray ["a", "b"])           == "(ab)"
+  assert $ foldr1 (\l r -> "(" <> l <> r <> ")") (fromArray ["a"])                == "a"
+
+  log "foldl1 should work"
+  assert $ foldl1 (\l r -> "(" <> l <> r <> ")") (fromArray ["a", "b", "c", "d"]) == "(a(b(cd)))"
```
JordanMartinez, line 313: Pretty sure this should be `foldr1`'s test:
```
"( <> "a" <>
  "(" <> "b" <>
    "(" <> "c" <>
      "(" <> "d" <>
      ")"
    ")"
  ")"
")"
```

JordanMartinez (changes requested):
I'm pretty sure you've swapped `foldr1` and `foldl1` implementations around.

src/Data/Array/NonEmpty/Internal.js
srghma: yes, I changed a name, because this name is better

JordanMartinez: Let me clarify. The function name is `foldr1Impl`, but the implementation is what I would define for `foldl1Impl`. My issue isn't that the you changed the name. My issue is that the name refers to the wrong implementation. If you changed this to `foldl1Impl` and changed the currently named `foldl1Impl` to `foldr1Impl`, then the names would refer to the correct functions. You would also need to update the tests, since they are referring to the wrong outputs as well.

test/Test/Data/Array/NonEmpty.purs
srghma: indeed, tnx, but the innermost d is not wrapped

the test is taken from https://github.com/purescript/purescript-nonempty/pull/39/files

src/Data/Array/NonEmpty/Internal.js
srghma: yes, fixed

JordanMartinez (approved):

test/Test/Data/Array/NonEmpty.purs
JordanMartinez: Thanks! I updated my code.

srghma:
Merged

PR #184, opened 2020-11-19
Remove some usages of unsafeCoerce

kl0tl:
This pull request replace some usages of `unsafeCoerce`, not by `coerce` like I suggested in https://github.com/purescript/purescript-arrays/pull/181#issuecomment-716061771 but by matching on and applying the `NonEmptyArray` constructor itself as done in https://github.com/purescript/purescript-strings/pull/130. I also adapted the comments from the other pull request.

### DIFF
--- a/src/Data/Array/NonEmpty/Internal.purs
+++ b/src/Data/Array/NonEmpty/Internal.purs
@@ -1,4 +1,7 @@
-module Data.Array.NonEmpty.Internal (NonEmptyArray) where
+-- | This module exports the `NonEmptyArray` constructor, which can be
+-- | overlooked when searching for issues in one's code. See the
+-- | constructor's documentation for more information.
+module Data.Array.NonEmpty.Internal (NonEmptyArray(..)) where
 
 import Prelude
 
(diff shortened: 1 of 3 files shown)


src/Data/Array/NonEmpty/Internal.purs
```diff
@@ -1,4 +1,7 @@
-module Data.Array.NonEmpty.Internal (NonEmptyArray) where
+-- | This module exports the `NonEmptyArray` constructor, which can be
```
hdgarrood, line 1: I think it might be worth discouraging this module’s use a bit more, by adding a disclaimer that this module is not intended for public use and its API may change at any time, and a breaking change won’t be reflected in the version number?

kl0tl: What do you think if this, with some bold all caps and newlines to let the disclaimer breath:

Capture d’écran 2020-11-20 à 09 18 48

It could be nice of Pursuit to display a warning on internal modules, so we can have some CSS on top and we don’t have to synchronize them across packages.

hdgarrood (approved):
Looks good :+1:

thomashoneyman (approved):

PR #187, opened 2020-11-25
Migrate to GitHub Actions for CI

thomashoneyman:
Migrates to GitHub Actions for CI. This serves as a demonstration for other repositories in the PureScript core organization which have tests (unlike the [same change made for `semirings`](https://github.com/purescript/purescript-semirings/pull/17), as that library doesn't have tests).


thomashoneyman:
If desired, I can also add caching for NPM / Bower dependencies, as well as caching for the `output` directory for the built PureScript code.

hdgarrood (approved):
Thanks! I’d suggest not bothering with caching since it’s another thing to potentially go wrong and it’s not saving us a whole lot of time in this case.

thomashoneyman:
Backlinking to https://github.com/purescript/purescript/issues/3962

PR #188, opened 2020-11-27
Add intersperse with tests

### LINKED ISSUE #15: intersperse
Very useful functions for combining arrays of monoids (like strings)

JordanMartinez:
Fixes #15

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -124,7 +124,6 @@ import Control.Alternative (class Alternative)
 import Control.Lazy (class Lazy, defer)
 import Control.Monad.Rec.Class (class MonadRec, Step(..), tailRecM2)
 import Control.Monad.ST as ST
-import Control.Monad.ST.Internal as STI
 import Data.Array.NonEmpty.Internal (NonEmptyArray(..))
 import Data.Array.ST as STA
 import Data.Array.ST.Iterator as STAI
@@ -573,7 +572,7 @@ intersperse a arr = case length arr of
           let unsafeGetElem idx = unsafePartial (unsafeIndex arr idx)
           out <- STA.empty
           _ <- STA.push (unsafeGetElem 0) out
-          STI.for 1 len \idx -> do
+          ST.for 1 len \idx -> do
             _ <- STA.push a out
             void (STA.push (unsafeGetElem idx) out)
           pure out
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -557,14 +557,14 @@ alterAt i f xs = maybe Nothing go (xs !! i)
 -- | must have two or more elements for this operation to take effect.
 -- |
 -- | For example:
--- | ```
+-- | ```purescript
 -- | intersperse " " [ "a", "b" ] == [ "a", " ", "b" ]
 -- |
 -- | intersperse 0 [ 1, 2, 3, 4, 5 ] == [ 1, 0, 2, 0, 3, 0, 4, 0, 5 ]
 -- | ```
 -- | If the array has one or zero elements, the outputted array is the same
 -- | as the input array.
--- | ```
+-- | ```purescript
 -- | intersperse " " [] == []
 -- | intersperse " " ["a"] == ["a"]
 -- | ```
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -560,7 +560,7 @@ alterAt i f xs = maybe Nothing go (xs !! i)
 -- | intersperse 0 [ 1, 2, 3, 4, 5 ] == [ 1, 0, 2, 0, 3, 0, 4, 0, 5 ]
 -- | ```
 -- |
--- | If the array has one or zero elements, the input array is returned.
+-- | If the array has less than two elements, the input array is returned.
 -- | ```purescript
 -- | intersperse " " [] == []
 -- | intersperse " " ["a"] == ["a"]
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -570,7 +570,7 @@ alterAt i f xs = maybe Nothing go (xs !! i)
 -- | ```
 intersperse :: forall a. a -> Array a -> Array a
 intersperse a arr = case length arr of
-  len | len < 2 -> arr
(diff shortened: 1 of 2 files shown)


JordanMartinez:
Can I get a review here?

src/Data/Array.purs
```diff
@@ -551,6 +553,33 @@ alterAt i f xs = maybe Nothing go (xs !! i)
...
+-- | intersperse " " [ "a", "b" ] == [ "a", " ", "b" ]
+-- |
+-- | intersperse 0 [ 1, 2, 3, 4, 5 ] == [ 1, 0, 2, 0, 3, 0, 4, 0, 5 ]
+-- | ```
+-- | If the array has one or zero elements, the outputted array is the same
+-- | as the input array.
+-- | ```
+-- | intersperse " " [] == []
+-- | intersperse " " ["a"] == ["a"]
+-- | ```
```
thomashoneyman, line 570:
```suggestion
-- | Inserts the given element in between each element in the array. The array
-- | must have two or more elements for this operation to take effect.
-- |
-- | ```purs
-- | intersperse " " [ "a", "b" ] == [ "a", " ", "b" ]
-- | intersperse 0 [ 1, 2, 3, 4, 5 ] == [ 1, 0, 2, 0, 3, 0, 4, 0, 5 ]
-- | ```
-- |
-- | If the array has one or zero elements, the resulting array is the same
-- | as the input array.
-- |
-- | ```purs
-- | intersperse " " [] == []
-- | intersperse " " ["a"] == ["a"]
-- | ```
```

thomashoneyman: Some minor typo / code block fixes here ^

JordanMartinez: I don't think the `purescript` part after the three opening ticks does anything other than add noise to the docs (i.e.  the word "purescript" appears before the code block or as the first word in it or something).

src/Data/Array.purs
```diff
@@ -551,6 +553,32 @@ alterAt i f xs = maybe Nothing go (xs !! i)
...
+-- | Inserts the given element in between each element in the array. The array
+-- | must have two or more elements for this operation to take effect.
+-- |
+-- | ```
+-- | intersperse " " [ "a", "b" ] == [ "a", " ", "b" ]
+-- | intersperse 0 [ 1, 2, 3, 4, 5 ] == [ 1, 0, 2, 0, 3, 0, 4, 0, 5 ]
+-- | ```
+-- |
+-- | If the array has one or zero elements, the resulting array is a copy
+-- | of the input array.
```
JordanMartinez, line 565: I decided to use copy since it's not technically the "same" array.

hdgarrood: It actually will be the same array with this implementation. If you mutate one, you’ll see that change reflected in the other.

src/Data/Array.purs
thomashoneyman: It doesn't do anything in the docs at the moment, though it _could_ add syntax highlighting. The word "purescript" doesn't appear in the code block -- see: https://pursuit.purescript.org/packages/purescript-arrays/5.3.1/docs/Data.Array#v:reverse

src/Data/Array.purs
JordanMartinez: Ah... Shoot, that's right. Let me fix that.

JordanMartinez: Fixed in latest commits

src/Data/Array.purs
JordanMartinez: Ok. I've added it

JordanMartinez:
CI builds.

src/Data/Array.purs
hdgarrood: I think the previous implementation was preferable - there's no reason to make a copy here, since `Array` is already immutable. You'd have to use an escape hatch like `unsafeThaw` or the FFI to mutate an `Array`, but that's not something we need to defend against, because it can only happen through misuse of said escape hatches.

JordanMartinez: True.
Latest commit reverts that change

src/Data/Array.purs
```diff
@@ -551,6 +553,31 @@ alterAt i f xs = maybe Nothing go (xs !! i)
...
 
+-- | Inserts the given element in between each element in the array. The array
+-- | must have two or more elements for this operation to take effect.
+-- |
+-- | ```purescript
+-- | intersperse " " [ "a", "b" ] == [ "a", " ", "b" ]
+-- | intersperse 0 [ 1, 2, 3, 4, 5 ] == [ 1, 0, 2, 0, 3, 0, 4, 0, 5 ]
+-- | ```
+-- |
+-- | If the array has one or zero elements, the input array is returned.
```
hdgarrood, line 564: I think "less than two" might read a bit more naturally here? Up to you.

src/Data/Array.purs
```diff
@@ -551,6 +553,31 @@ alterAt i f xs = maybe Nothing go (xs !! i)
...
+-- | intersperse " " ["a"] == ["a"]
+-- | ```
+intersperse :: forall a. a -> Array a -> Array a
+intersperse a arr = case length arr of
+  len | len < 2 -> arr
+      | otherwise -> STA.run do
+          let unsafeGetElem idx = unsafePartial (unsafeIndex arr idx)
+          out <- STA.empty
+          _ <- STA.push (unsafeGetElem 0) out
+          STI.for 1 len \idx -> do
```
hdgarrood, line 576: I think `for` is re-exported from `Control.Monad.ST`, so I don't think we need to import the Internal module here (and if we can avoid importing the Internal module, it would be good to do that).

hdgarrood (approved):
Approving because I think this is good to merge once these comments have been addressed. Thanks! :+1:

src/Data/Array.purs
JordanMartinez: Ah! Good catch! Fixed in latest commits.

src/Data/Array.purs
JordanMartinez: I don't know. I think I could go either way. 
Meh, let's just use the "less than two."

thomashoneyman (approved):

PR #189, opened 2020-11-28
Add short circuiting functions of `elem`, `notElem`, `find`, `findMap`, `scanl`, and `scanr`

JordanMartinez:
See https://github.com/purescript/purescript-arrays/pull/173#issuecomment-734956092

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -135,7 +135,6 @@ import Data.Array.ST.Iterator as STAI
 import Data.Foldable (class Foldable, foldl, foldr, traverse_)
 import Data.Foldable (foldl, foldr, foldMap, fold, intercalate, any, all) as Exports
 import Data.Maybe (Maybe(..), maybe, isJust, fromJust, isNothing)
-import Data.Traversable (scanl, scanr) as Exports
 import Data.Traversable (sequence, traverse)
 import Data.Tuple (Tuple(..), fst, snd)
 import Data.Unfoldable (class Unfoldable, unfoldr)
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -57,6 +57,7 @@ module Data.Array
   , elemIndex
   , elemLastIndex
   , find
+  , findMap
   , findIndex
   , findLastIndex
   , insertAt
@@ -130,7 +131,7 @@ import Data.Array.NonEmpty.Internal (NonEmptyArray(..))
 import Data.Array.ST as STA
 import Data.Array.ST.Iterator as STAI
 import Data.Foldable (class Foldable, foldl, foldr, traverse_)
-import Data.Foldable (foldl, foldr, foldMap, fold, intercalate, findMap, any, all) as Exports
+import Data.Foldable (foldl, foldr, foldMap, fold, intercalate, any, all) as Exports
 import Data.Maybe (Maybe(..), maybe, isJust, fromJust, isNothing)
 import Data.Traversable (scanl, scanr) as Exports
 import Data.Traversable (sequence, traverse)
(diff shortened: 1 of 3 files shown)


JordanMartinez:
Can I get a review on this?

test/Test/Data/Array.purs
```diff
@@ -139,6 +147,14 @@ testArray = do
   assert $ (A.elemLastIndex 1 [1, 2, 1]) == Just 2
   assert $ (A.elemLastIndex 4 [1, 2, 1]) == Nothing
 
+  log "find should return the first element for which a predicate returns true in an array"
+  assert $ (A.find (_ /= 1) [1, 2, 1]) == Just 2
+  assert $ (A.find (_ == 3) [1, 2, 1]) == Nothing
+
+  log "findMap should return the mapping of the first element that satisfies the given predicate"
```
hdgarrood, line 154: I think it would be good to also add a test case for an array where there is more than one element which satisfies the predicate here.

test/Test/Data/Array.purs
```diff
@@ -237,6 +253,14 @@ testArray = do
   assert $ A.modifyAtIndices [0, 2, 8] not [true,  true, true,  true] ==
                                            [false, true, false, true]
 
+  log "scanl should return an array that stores the accumulated value at each step"
+  assert $ A.scanl (+)  0 [1,2,3] == [1, 3, 6]
```
hdgarrood, line 257: Can we test that these agree with the Foldable versions please? It's probably not necessary to compare with the Foldable versions for the other functions, but this one differs from Haskell and so there's more than one "sensible" option for how to implement it, so I think it's worth being a little bit more careful/deliberate in the tests.

src/Data/Array.purs
```diff
@@ -698,6 +733,32 @@ modifyAtIndices :: forall t a. Foldable t => t Int -> (a -> a) -> Array a -> Arr
...
+-- | appear in the result (unlike Haskell's `Prelude.scanl`).
+-- |
+-- | ```
+-- | scanl (+) 0  [1,2,3] = [1,3,6]
+-- | scanl (-) 10 [1,2,3] = [9,7,4]
+-- | ```
+scanl :: forall a b. (b -> a -> b) -> b -> Array a -> Array b
+scanl = scanlImpl
+
+foreign import scanlImpl :: forall a b. (b -> a -> b) -> b -> Array a -> Array b
```
hdgarrood, line 747: Since the FFI implementation is ready to use as-is, we can just import it as `scanl` straight away; I don't think there's any need to alias it.

JordanMartinez: Fixed in latest commits

test/Test/Data/Array.purs
JordanMartinez: Fixed in latest commits

test/Test/Data/Array.purs
JordanMartinez: Could you clarify what you mean by this? The tests I used were taken from the [`scanl` docs](https://pursuit.purescript.org/packages/purescript-foldable-traversable/4.1.1/docs/Data.Traversable#v:scanl) and [`scanr` docs](https://pursuit.purescript.org/packages/purescript-foldable-traversable/4.1.1/docs/Data.Traversable#v:scanr)

test/Test/Data/Array.purs
hdgarrood: Sorry, I should have been clearer: I think we should have a test which rules out the possibility that our `findMap` actually returns the _last_ match.

JordanMartinez: I've updated the tests to account for that.

test/Test/Data/Array.purs
JordanMartinez: Rereading this again, I think you mean this:
```purescript
assert $ A.scanl (+)  0 [1,2,3] == scanl (+) 0 [1,2,3]
```

JordanMartinez:
I believe I've addressed all your comments.

test/Test/Data/Array.purs
hdgarrood: Yeah, exactly.

hdgarrood (approved):
LGTM, thanks

Issue #186, opened 2020-11-24
Rename empty to new in Data.Array.ST


kl0tl:
The foreign object equivalent of Data.Array.ST.empty is called new and in https://github.com/purescript/purescript-record/issues/68 I proposed to add Record.ST.new for consistency. Should we also rename Data.Array.ST.empty to new then?

JordanMartinez:
I think we should be consistent across all of the ST repos, so I'm for this.

JordanMartinez:
This was closed by #191.

PR #191, opened 2020-12-13
Rename Data.Array.ST.empty

kl0tl:
This pull request deprecate Data.Array.ST.empty and alias it to Data.Array.ST.new, so that it can be removed in a future release, for consistency with Foreign.Object.ST.new (see https://github.com/purescript/purescript-arrays/issues/186).


thomashoneyman:
I think we’re defaulting to a deprecation strategy whenever it isn’t too onerous, perhaps with an accompanying issue to remove deprecations in the next cycle

JordanMartinez (approved):

PR #193, opened 2020-12-19
Export specialized any and all from Data.Array and Data.Array.NonEmpty

kl0tl:
Following https://github.com/purescript/purescript-arrays/pull/189, this pull request specialize two more functions. Exporting `any, all :: forall a. (a -> Boolean) -> Array a -> Boolean` instead of `any, all :: forall a b f. Foldable f => HeytingAlgebra b => (a -> b) -> f a -> b` allows to stop iterating the array as soon as the predicate yields true for any or false for all.


hdgarrood (approved):

src/Data/Array.purs
```diff
@@ -1192,6 +1195,12 @@ unzip xs =
     snds' <- STA.unsafeFreeze snds
     pure $ Tuple fsts' snds'
 
+-- | Returns true if at least one array element satisfy the given predicate.
```
hdgarrood, line 1198: We want “satisfies” here rather than “satisfy” for subject-verb agreement. Also, I think it would be nice to explain why we have specialised HeytingAlgebra to Boolean?

hdgarrood: A couple of simple examples would also be nice, I guess

thomashoneyman (approved):
Agree with Harry's comments, and otherwise this looks quite good. 👍

src/Data/Array.purs
kl0tl: I updated the documentation, added some examples I fancied more than the tests and updated the tests.

PR #195, opened 2020-12-21
Sort undefined according to the comparison function

kl0tl:
This pull request implements Data.Array.sortBy and Data.Array.ST.sortBy with the merge sort proposed in https://github.com/purescript/purescript-arrays/issues/155, so that `undefined` is sorted according to the comparison function and not to the end of the array (this also applies to Data.Array.NonEmpty.sortBy because it is implemented with Data.Array.sortBy). The only liberty I took with the implementation was to get rid of the uncurried functions since we’re not using any elsewhere in this library.

Do we also want `unsafeSort :: forall a. Ord a => Array a -> Array a`, `unsafeSortBy :: forall a. (a -> a -> Ordering) -> Array a -> Array a` and `unsafeSortWith :: forall a b. Ord b => (a -> b) -> Array a -> Array a` functions, implemented with the native Array.prototype.sort method?

Close https://github.com/purescript/purescript-arrays/issues/155.

### DIFF
--- a/src/Data/Array.js
+++ b/src/Data/Array.js
@@ -270,13 +270,65 @@ exports.scanr = function (f) {
 // Sorting ---------------------------------------------------------------------
 //------------------------------------------------------------------------------
 
-exports.sortImpl = function (f) {
-  return function (l) {
-    return l.slice().sort(function (x, y) {
-      return f(x)(y);
-    });
+exports.sortByImpl = (function () {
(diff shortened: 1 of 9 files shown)


hdgarrood (approved):
Looks good! I wouldn’t bother with `unsafeSort`, I expect it’s quite rare that there’s a difference that matters and it’ll be a pain for alternate backend maintainers.

PR #164, opened 2020-01-16
Add groupBy to Array.NonEmpty

Zelenya:
Add `groupBy`, `group`, `group'` to `Array.NonEmpty`

PR #200, opened 2020-12-28
Rename group' to groupAll and add groupAllBy; add docs

JordanMartinez:
Addresses the comments that @kl0tli made [here](https://github.com/purescript/purescript-arrays/pull/164#issuecomment-750843053)﻿

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -959,7 +959,7 @@ group xs = groupBy eq xs
 -- | groupAll [1, 1, 2, 2, 1] == [NonEmptyArray [1, 1, 1], NonEmptyArray [2, 2]]
 -- | ```
 groupAll :: forall a. Ord a => Array a -> Array (NonEmptyArray a)
-groupAll = groupAllBy eq
+groupAll = groupAllBy compare
 
 -- | Deprecated previous name of `groupAll`.
 group' :: forall a. Warn (Text "'group\'' is deprecated, use 'groupAll' instead") => Ord a => Array a -> Array (NonEmptyArray a)
@@ -987,15 +987,15 @@ groupBy op xs =
     STA.unsafeFreeze result
 
 -- | Group equal elements of an array into arrays, using the specified
--- | equivalence relation to determine equality.
+-- | comparison relation to determine equality.
 -- |
 -- | ```purescript
--- | groupAllBy (\a b -> odd a && odd b) [1, 3, 2, 4, 3, 3]
--- |    = [NonEmptyArray [1], NonEmptyArray [2], NonEmptyArray [3, 3, 3], NonEmptyArray [4]]
+-- | groupAllBy (\a b -> compare (odd a) (odd b)) [1, 3, 2, 4, 3, 3]
+-- |    = [NonEmptyArray [2, 4], NonEmptyArray [1, 3, 3, 3]]
 -- | ```
 -- |
-groupAllBy :: forall a. Ord a => (a -> a -> Boolean) -> Array a -> Array (NonEmptyArray a)
-groupAllBy p = groupBy p <<< sort
+groupAllBy :: forall a. (a -> a -> Ordering) -> Array a -> Array (NonEmptyArray a)
+groupAllBy cmp = groupBy (\x y -> cmp x y == EQ) <<< sortBy cmp
 
 -- | Remove the duplicates from an array, creating a new array.
 -- |
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -987,11 +987,11 @@ groupBy op xs =
     STA.unsafeFreeze result
 
 -- | Group equal elements of an array into arrays, using the specified
--- | comparison relation to determine equality.
+-- | partial ordering relation to determine equality.
 -- |
 -- | ```purescript
--- | groupAllBy (\a b -> compare (odd a) (odd b)) [1, 3, 2, 4, 3, 3]
--- |    = [NonEmptyArray [2, 4], NonEmptyArray [1, 3, 3, 3]]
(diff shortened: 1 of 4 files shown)


JordanMartinez:
CI passes. Ready for review.

hdgarrood:
This applies to #194 too but I wonder if it's worth considering dropping the `Ord` constraint in groupAllBy and instead having it take an `a -> a -> Ordering` to use for both the sorting and the grouping.

JordanMartinez:
Perhaps a variant called `groupAllBy'`?
```purescript
groupAllBy' :: forall a. (a -> a -> Boolean) -> (a -> a -> Ordering) -> Array a -> Array (NonEmptyArray a)
groupAllBy' p c = groupBy p <<< sortBy c
```
Does this pass the Fairbairn threshold?

src/Data/Array/NonEmpty.purs
```diff
@@ -350,15 +353,49 @@ span
...
+-- |
+-- | ```purescript
+-- | groupAll (NonEmptyArray [1, 1, 2, 2, 1]) ==
+-- |   NonEmptyArray [NonEmptyArray [1, 1, 1], NonEmptyArray [2, 2]]
+-- | `
+groupAll :: forall a. Ord a => NonEmptyArray a -> NonEmptyArray (NonEmptyArray a)
+groupAll = groupAllBy eq
+
+-- | Deprecated previous name of `groupAll`.
+group' :: forall a. Warn (Text "'group\'' is deprecated, use groupAll instead") => Ord a => NonEmptyArray a -> NonEmptyArray (NonEmptyArray a)
```
kl0tl, line 375: Looks like I forgot to quote groupAll in the deprecation warning of Data.Array.group'. Could you do it there and here?

kl0tl (commented):
You’re right Harry, I think we should update Data.Array.groupAllBy to `forall a. (a -> a -> Ordering) -> Array a -> Array (NonEmptyArray a)` (ditto for the non empty variant).

src/Data/Array/NonEmpty.purs
JordanMartinez: Sure!

JordanMartinez: Done

JordanMartinez:
> `forall a. (a -> a -> Ordering) -> Array a -> Array (NonEmptyArray a)`

Meaning this?
```purescript
groupAllBy :: forall a. (a -> a -> Ordering) -> Array a -> Array (NonEmptyArray a)
groupAllBy cmp = groupBy (eq Eq $ cmp) <<< sortBy cmp
```

kl0tl:
Something like the following should work:

```purs
groupAllBy comp = groupBy (\x y -> comp x y == EQ) <<< sortBy comp
```

src/Data/Array.purs
```diff
@@ -987,15 +987,15 @@ groupBy op xs =
...
 -- | Group equal elements of an array into arrays, using the specified
--- | equivalence relation to determine equality.
+-- | comparison relation to determine equality.
 -- |
 -- | ```purescript
--- | groupAllBy (\a b -> odd a && odd b) [1, 3, 2, 4, 3, 3]
--- |    = [NonEmptyArray [1], NonEmptyArray [2], NonEmptyArray [3, 3, 3], NonEmptyArray [4]]
+-- | groupAllBy (\a b -> compare (odd a) (odd b)) [1, 3, 2, 4, 3, 3]
+-- |    = [NonEmptyArray [2, 4], NonEmptyArray [1, 3, 3, 3]]
 -- | ```
```
JordanMartinez, line 995: I adapted the example to the one above, but I don't think it's a good example. Any ideas?

src/Data/Array.purs
```diff
@@ -987,15 +987,15 @@ groupBy op xs =
     STA.unsafeFreeze result
 
 -- | Group equal elements of an array into arrays, using the specified
--- | equivalence relation to determine equality.
+-- | comparison relation to determine equality.
 -- |
```
JordanMartinez, line 991: How should docs be updated to better communicate what this does?

kl0tl: I think we could replace “comparison relation” by “partial ordering“, as in the documentation of sortBy, here.

src/Data/Array.purs
kl0tl: What about using `comparing Down`?

```purs
groupAllBy (comparing Down) [1, 3, 2, 4, 3, 3]
  = [NonEmptyArray [4], NonEmptyArray [3, 3, 3], NonEmptyArray [2], NonEmptyArray [1]]
```

test/Test/Data/Array.purs
```diff
@@ -363,10 +363,10 @@ testArray = do
   assert $ A.groupBy (\_ _ -> true) [1, 2, 3] == [nea [1, 2, 3]]
 
   log "groupAllBy should group equal elements into arrays based on an equivalence relation"
```
kl0tl, line 365: What about “based on the result of a comparison function”, as in the test for sortBy?

src/Data/Array.purs
hdgarrood: Those docs are actually incorrect - it's expecting a total ordering, not a partial one. I think they must be an oversight left over from a short period in the distant past where we said Ord represented a partial ordering rather than a total ordering. We call this a "comparison function" in Data.Array.ST, and we say "where elements are compared according to the specified ordering" in Data.List. I think either of those are fine.

JordanMartinez:
I've addresses all your comments.

src/Data/Array.purs
JordanMartinez: I've reverted back to using `comparison function`

JordanMartinez:
This is ready for another look over.

hdgarrood (approved):

PR #201, opened 2020-12-28
Add missing members to Data.Array.NonEmpty

kl0tl:
We added a few functions to Data.Array in https://github.com/purescript/purescript-arrays/pull/189, https://github.com/purescript/purescript-arrays/pull/188 and https://github.com/purescript/purescript-arrays/pull/68 but forgot to export variants from Data.Array.NonEmpty.

This pull request adapts them to non empty arrays and also re-exports foldl1, foldr1, foldMap1, fold1 and intercalate (from Data.Semigroup.Foldable) so that Data.Array and Data.Array.NonEmpty have (mostly) the same interface.

### DIFF
--- a/test/Test/Data/Array/NonEmpty.purs
+++ b/test/Test/Data/Array/NonEmpty.purs
@@ -6,7 +6,7 @@ import Data.Array as A
 import Data.Array.NonEmpty as NEA
 import Data.Const (Const(..))
 import Data.Foldable (for_, sum, traverse_)
-import Data.FunctorWithIndex (mapWithIndex)
+import Data.Traversable (scanl, scanr)
 import Data.Maybe (Maybe(..), fromJust)
 import Data.Monoid.Additive (Additive(..))
 import Data.NonEmpty ((:|))
@@ -98,6 +98,14 @@ testNonEmptyArray = do
   assert $ NEA.index (fromArray [1, 2, 3]) 6 == Nothing
   assert $ NEA.index (fromArray [1, 2, 3]) (-1) == Nothing
 
+  log "elem should return true if the array contains the given element at least once"
+  assert $ NEA.elem 1 (fromArray [1, 2, 1]) == true
+  assert $ NEA.elem 4 (fromArray [1, 2, 1]) == false
+
+  log "notElem should return true if the array does not contain the given element"
+  assert $ NEA.notElem 1 (fromArray [1, 2, 1]) == false
+  assert $ NEA.notElem 4 (fromArray [1, 2, 1]) == true
+
   log "elemIndex should return the index of an item that a predicate returns true for in an array"
   assert $ NEA.elemIndex 1 (fromArray [1, 2, 1]) == Just 0
   assert $ NEA.elemIndex 4 (fromArray [1, 2, 1]) == Nothing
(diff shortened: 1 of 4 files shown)


hdgarrood:
I think we should export monomorphic versions of the fold* functions in both this module and in Data.Array, see https://github.com/purescript/purescript-arrays/pull/173.

JordanMartinez:
Data.Array.NonEmpty does not have `groupAll` or `groupAllBy`. I'm assuming we want these included there, too, correct?

JordanMartinez:
I've updated `Data.Array` and `Data.Array.NonEmpty` to export monomorphic versions of `foldl`, `foldr`, `fold1`, `foldMap`, and `intercalate`.

JordanMartinez:
> Data.Array.NonEmpty does not have groupAll or groupAllBy. I'm assuming we want these included there, too, correct?

Nevermind. This specific PR doesn't include those members in its exports because it was based on a previous branch. Once I merged it with `master`, that was no longer the case.

test/Test/Data/Array/NonEmpty.purs
```diff
@@ -106,6 +114,15 @@ testNonEmptyArray = do
   assert $ NEA.elemLastIndex 1 (fromArray [1, 2, 1]) == Just 2
   assert $ NEA.elemLastIndex 4 (fromArray [1, 2, 1]) == Nothing
 
+  log "find should return the first element for which a predicate returns true in an array"
+  assert $ NEA.find (_ /= 1) (fromArray [1, 2, 1]) == Just 2
+  assert $ NEA.find (_ == 3) (fromArray [1, 2, 1]) == Nothing
```
JordanMartinez, line 123: Should there be another test here to verify that the first element is returned when two or more elements would satisfy the predicate?

JordanMartinez (approved):

kl0tl:
Shouldn’t we rather remove the re-exports? That’s quite inconsistent with https://github.com/purescript/purescript-nonempty/pull/42, where we deprecated the monomorphic Data.NonEmpty.foldl1, otherwise.

test/Test/Data/Array/NonEmpty.purs
kl0tl: I think so. We could also ensure that findIndex actually returns the index of the first match by changing the `_ /= 1` predicate to `_ == 1`  .

hdgarrood:
I suppose it is but if we want these to be consistent then we should reintroduce the NonEmpty ones. Removing these re-exports from Data.Array is much too breaking to be worthy of consideration in my mind.

JordanMartinez:
> Shouldn’t we rather remove the re-exports?

I think that was deprecated because, while the `NonEmpty` part is monomorphic, its underlying `f` is still polymorphic. Here, we have a monomorphic `Array` or `NonEmptyArray`.

test/Test/Data/Array/NonEmpty.purs
JordanMartinez: Fixed

hdgarrood (approved):

JordanMartinez:
I'll open an PR that re-adds the monomorphic `fold1` to NonEmpty

JordanMartinez:
See https://github.com/purescript/purescript-nonempty/pull/45

PR #203, opened 2020-12-31
Make `union*`, `difference*`, and `intersect*` run linearly by sorting elements first

### LINKED ISSUE #192: Use Ord by default for all set-like operations
We already made `nub` use Ord by default in #91, but there are a few other unnecessarily quadratic functions in this library still which probably deserve the same treatment. I think these are `union`, `intersect`, and `difference`.

As well as being quadratic, `union` and `difference` have a strange-seeming behaviour where duplicates are only preserved in the first argument:
```purescript
union [0, 0] [1, 1] == [0, 0, 1]
```
See #56 for more info. We ought to be careful about changing this behaviour but I think we should consider it, since I think it’s very likely to trip people up and we will be making breaking changes to these functions anyway if we do this.

JordanMartinez:
Fixes #192

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -1087,7 +1087,7 @@ union = unionBy compare
 -- | ```
 -- |
 unionBy :: forall a. (a -> a -> Ordering) -> Array a -> Array a -> Array a
-unionBy cmp left right = map snd $ sortWith fst $ ST.run do
+unionBy cmp left right = map snd $ ST.run do
   result <- STA.new
   ST.foreach indexedAndSorted \(Tuple fromLeftArray pair@(Tuple _ x')) -> do
     if fromLeftArray then do
@@ -1100,6 +1100,7 @@ unionBy cmp left right = map snd $ sortWith fst $ ST.run do
           | otherwise -> pure unit
         Nothing -> do
           void $ STA.push pair result
+  _ <- STA.sortWith fst result
   STA.unsafeFreeze result
   where
     -- Note: when elements are equal, left array elements
@@ -1155,7 +1156,7 @@ infix 5 difference as \\
 differenceBy :: forall a. (a -> a -> Ordering) -> Array a -> Array a -> Array a
 differenceBy   _ left       [] = left
 differenceBy   _ left@[]     _ = left
-differenceBy cmp left    right = map snd $ sortWith fst $ ST.run do
+differenceBy cmp left    right = map snd $ ST.run do
   result <- STA.new
   latestRightArrayValue <- STRef.new Nothing
   ST.foreach indexedAndSorted \(Tuple fromLeftArray pair@(Tuple _ x')) -> do
@@ -1177,6 +1178,7 @@ differenceBy cmp left    right = map snd $ sortWith fst $ ST.run do
         void $ STRef.write next latestRightArrayValue
       false, _ -> do
         void $ STRef.write (Just (Tuple x' 1)) latestRightArrayValue
+  _ <- STA.sortWith fst result
   STA.unsafeFreeze result
   where
     -- Note: when elements are equal, right array elements
@@ -1244,7 +1246,7 @@ intersect = intersectBy compare
 intersectBy :: forall a. (a -> a -> Ordering) -> Array a -> Array a -> Array a
 intersectBy _   left       [] = left
 intersectBy _   left@[]     _ = left
-intersectBy cmp left    right = map snd $ sortWith fst $ ST.run do
+intersectBy cmp left    right = map snd $ ST.run do
   result <- STA.new
   latestRightArrayValue <- STRef.new Nothing
   ST.foreach indexedAndSorted \(Tuple fromLeftArray pair@(Tuple i x')) ->
@@ -1256,6 +1258,7 @@ intersectBy cmp left    right = map snd $ sortWith fst $ ST.run do
         _ -> pure unit
     else do
       void $ STRef.write (Just x') latestRightArrayValue
+  _ <- STA.sortWith fst result
   STA.unsafeFreeze result
   where
     -- Note: when elements are equal, right array elements
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -143,6 +143,7 @@ import Data.Array.ST.Iterator as STAI
 import Data.Foldable (class Foldable, foldl, foldr, traverse_)
 import Data.Foldable (foldl, foldr, foldMap, fold, intercalate) as Exports
 import Data.Maybe (Maybe(..), maybe, isJust, fromJust, isNothing)
+import Data.Ordering (invert)
 import Data.Traversable (sequence, traverse)
 import Data.Tuple (Tuple(..), fst, snd)
 import Data.Unfoldable (class Unfoldable, unfoldr)
@@ -1073,8 +1074,8 @@ nubByEq eq xs = ST.run do
 -- | union [1, 2, 1, 1] [3, 3, 3, 4] = [1, 2, 1, 1, 3, 4]
 -- | ```
 -- |
-union :: forall a. Eq a => Array a -> Array a -> Array a
-union = unionBy (==)
+union :: forall a. Ord a => Array a -> Array a -> Array a
+union = unionBy compare
 
 -- | Calculate the union of two arrays, using the specified function to
 -- | determine equality of elements. Note that duplicates in the first array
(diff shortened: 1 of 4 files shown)


JordanMartinez:
The failing test is:
```purescript
log "(\\\\) should return the difference between two lists"
assert $ [1, 2, 3, 4, 3, 2, 1] \\ [1, 1, 2, 3] == [4, 3, 2]
```
My algorithm will return `[4]` rather than `[4, 3, 2]`.

JordanMartinez:
I've updated the `intersectionBy` and `difference`/`differenceBy` to work and pass CI. I'm still not sure how `union` differs from `alt`/`<|>`, so I'm going to wait until that question is clarified.

JordanMartinez:
This PR is ready for review

hdgarrood:
I haven't reviewed properly yet but just so you know, sorting based on a comparison function which only compares two elements at a time is always at least O(n log n), which is beyond linear. Not by that much, admittedly, since `log n` grows very slowly, but it's incorrect to say that it is linear if we're sorting.

JordanMartinez:
Hmm..... Besides the `eq` to `compare` change and the lack of `STFn`, is there anything else that might make this benchmark misleading? Everything is slower by quite a bit.

Assume the following values for each benchmark below

```purescript
let 
  shortNats = Array.range 0 100
  longNats = Array.range 0 10000
  mod3Eq x y = (x `mod` 3) == (y `mod` 3)
  mod3Cmp x y = compare (x `mod` 3) (y `mod` 3)
  
-- `master` uses `mod3Eq` whereas this PR uses `mod3Cmp`
```

`unionBy`
  
```purescript
benchUnionBy = do
    log $ "unionBy (" <> show (Array.length shortNats) <> ")"
    benchWith 1000 \_ -> Array.unionBy mod3Eq shortNats shortNats

    log $ "unionBy (" <> show (Array.length longNats) <> ")"
    benchWith 100 \_ -> Array.unionBy mod3Eq longNats longNats
```

```
unionBy -- master branch
---------------
unionBy (101)
mean   = 60.73 μs
stddev = 81.23 μs
min    = 34.67 μs
max    = 1.36 ms
unionBy (10001)
mean   = 3.40 ms
stddev = 1.56 ms
min    = 2.98 ms
max    = 18.20 ms

unionBy -- this PR
---------------
unionBy (101)
mean   = 208.75 μs
stddev = 197.98 μs
min    = 162.88 μs
max    = 3.23 ms
unionBy (10001)
mean   = 32.38 ms
stddev = 6.30 ms
min    = 28.67 ms
max    = 73.98 ms
```

`intersectBy`

```purescript
  benchIntersectBy = do
    log $ "intersectBy (" <> show (Array.length shortNats) <> ")"
    benchWith 1000 \_ -> Array.intersectBy mod3Eq shortNats shortNats

    log $ "intersectBy (" <> show (Array.length longNats) <> ")"
    benchWith 100 \_ -> Array.intersectBy mod3Eq longNats longNats
```

```
intersectBy -- current master
---------------
intersectBy (101)
mean   = 16.93 μs
stddev = 39.69 μs
min    = 7.69 μs
max    = 789.20 μs
intersectBy (10001)
mean   = 760.82 μs
stddev = 549.92 μs
min    = 626.33 μs
max    = 6.13 ms

intersectBy -- this PR
---------------
intersectBy (101)
mean   = 205.14 μs
stddev = 150.04 μs
min    = 158.94 μs
max    = 2.21 ms
intersectBy (10001)
mean   = 31.81 ms
stddev = 5.34 ms
min    = 28.55 ms
max    = 62.24 ms
```

`difference`

```purescript
  benchDifference = do
    log $ "difference (" <> show (Array.length shortNats) <> ")"
    benchWith 1000 \_ -> Array.difference shortNats shortNats

    log $ "difference (" <> show (Array.length longNats) <> ")"
    benchWith 100 \_ -> Array.difference longNats longNats
```

```
difference -- current master
---------------
difference (101)
mean   = 79.57 μs
stddev = 48.39 μs
min    = 63.89 μs
max    = 742.35 μs
difference (10001)
mean   = 492.35 ms
stddev = 24.31 ms
min    = 463.04 ms
max    = 594.56 ms

difference -- this PR
---------------
difference (101)
mean   = 142.91 μs
stddev = 136.28 μs
min    = 106.35 μs
max    = 1.73 ms
difference (10001)
mean   = 21.30 ms
stddev = 3.75 ms
min    = 18.21 ms
max    = 44.52 ms
```

src/Data/Array.purs
```diff
@@ -1118,13 +1146,82 @@ deleteBy eq x ys = maybe ys (\i -> unsafePartial $ fromJust (deleteAt i ys)) (fi
...
+--   rightIndexedPlus = mapWithIndex tupleAdjusted right
+--
+-- in leftIndexed <> rightIndexedPlus
+-- ```
+-- ... but without creating two intermediate arrays due to the `mapWithIndex`
+-- on both arrays.
+-- ```
+-- let t3 a b c = Tuple a (Tuple b c)
+-- combineIndex compare [0] [0, 1]
+--   == [t3 true 0 0, t3 false 1 0, t3 false 2 1]
```
milesfrain, line 1210:
```suggestion
-- combineIndex compare [7] [9, 5]
--   == [t3 true 0 7, t3 false 1 9, t3 false 2 5]
```
Easier to follow example with values that are distinct from indices

src/Data/Array.purs
```diff
@@ -1118,13 +1146,82 @@ deleteBy eq x ys = maybe ys (\i -> unsafePartial $ fromJust (deleteAt i ys)) (fi
...
+-- ```
+combineIndex :: forall a. Array a -> Array a -> Array (Tuple Boolean (Tuple Int a))
+combineIndex left right = ST.run do
+  out <- STA.new
```
milesfrain, line 1214: Are there any benefits to preallocating the array to a known size? Tried to find a clear answer for this in JS. And if there is a benefit that we want to take advantage of, we'd need to expand the `Array.ST` API.

src/Data/Array.purs
```diff
@@ -1118,13 +1146,82 @@ deleteBy eq x ys = maybe ys (\i -> unsafePartial $ fromJust (deleteAt i ys)) (fi
...
+-- Essentially...
+-- ```
+-- combineWithIndex cmp left right =
+-- let
+--   leftIndexed = mapWithIndex (\idx v -> Tuple idx v) left
+--
+--   tupleAdjusted idx v = Tuple (idx + (length left)) v
+--   rightIndexedPlus = mapWithIndex tupleAdjusted right
+--
+-- in leftIndexed <> rightIndexedPlus
```
milesfrain, line 1203: Looks like this is missing the part where it sets the boolean flags

src/Data/Array.purs
```diff
@@ -1118,13 +1146,82 @@ deleteBy eq x ys = maybe ys (\i -> unsafePartial $ fromJust (deleteAt i ys)) (fi
...
+combineIndex left right = ST.run do
+  out <- STA.new
+  let leftLen = length left
+  ST.for 0 leftLen \idx -> do
+    let val = unsafePartial $ unsafeIndex left idx
+    void $ STA.push (Tuple true (Tuple idx val)) out
+  let rightLen = length right
+  ST.for 0 rightLen \idx -> do
+    let val = unsafePartial $ unsafeIndex right idx
+    void $ STA.push (Tuple false (Tuple (leftLen + idx) val)) out
```
milesfrain, line 1222: Are there any performance penalties to using a record instead of a tuple? Luckily, this tuple happens to be type-safe because of the distinct types (`Boolean`, `Int`, `a`).

src/Data/Array.purs
```diff
@@ -1083,8 +1086,33 @@ union = unionBy (==)
 -- | unionBy mod3eq [1, 5, 1, 2] [3, 4, 3, 3] = [1, 5, 1, 2, 3]
 -- | ```
 -- |
-unionBy :: forall a. (a -> a -> Boolean) -> Array a -> Array a -> Array a
-unionBy eq xs ys = xs <> foldl (flip (deleteBy eq)) (nubByEq eq ys) xs
+unionBy :: forall a. (a -> a -> Ordering) -> Array a -> Array a -> Array a
+unionBy cmp left right = map snd $ ST.run do
```
milesfrain, line 1090: It may be more efficient to take advantage of the of the fact that the output is always the first array plus some other stuff appended to it. Here's some non-ST pseudo-ish code describing that option:
```purs
union left right = left <> other where
  other =
    combineIndex right left -- intentionally putting right first and assuming stable sort
    # sortBy valueFunc
    # groupBy valueFunc
    # map head
    # filter keepRightOnly
```

src/Data/Array.purs
JordanMartinez: But then there aren't values that exist in both arrays.

src/Data/Array.purs
JordanMartinez: I've wondered about that myself...

src/Data/Array.purs
JordanMartinez: Does it need to?

src/Data/Array.purs
JordanMartinez: Not sure. I also tried it out with `data Tuple3 a b c = Tuple3 a b c` and that sped things up only slightly.

src/Data/Array.purs
JordanMartinez: What would be the runtime cost of this approach versus mine?

I made the rather stupid assumption that mine would be `n+m` because when Harry said the current version was `n*m`, I thought he was implying we should make the code linear. I didn't actually analyze my code using Big O notation to see how many steps it takes.

src/Data/Array.purs
JordanMartinez: Nevermind. The point of that comment is to explain what's going on in the internals below more succinctly. I'll update this comment.

src/Data/Array.purs
milesfrain: Since `combineIndex` doesn't care about duplicates, I don't think we need the example to involve duplicates.
This would be like saying that `[1,2] <> [3,4] == [1,2,3,4]` is a bad example for append because it doesn't clarify that duplicates are preserved.

Could compromise with something like `[7] [7, 5]`.

src/Data/Array.purs
milesfrain: I believe all reasonable approaches are `O(n log n)` (where n is the sum of array sizes). But benchmarking could still reveal a 3x speedup with a different approach, which would still be `O(n log n)`. I'll play around with this and report back with findings.

hdgarrood:
I think these code changes should wait until we have reached a consensus on how these functions should actually work - there are still unknowns to do with things the handling of duplicates. I’m not sure how much it helps to change `union` to use Ord now if we plan to change its behaviour with respect to duplicates later, as that later change would still be breaking even if the type is the same.

JordanMartinez:
So, will this change be merged before `v0.14.0`? Or should we discuss it further and include it in `v0.15.0`?

src/Data/Array.purs
JordanMartinez: > Since `combineIndex` doesn't care about duplicates, I don't think we need the example to involve duplicates.

:man_facepalming: You're right! I forgot that this function doesn't need to care unlike the ones I'm working on. I'll make that fix.

JordanMartinez:
Minor optimization by only working with indices and copying values only once to the final array.

hdgarrood:
I’m tempted to leave it for 0.15.0 personally.

JordanMartinez:
I think this shouldn't be implemented until we have `STFn`. At the very least, the performance hit doesn't make it worth it in my mind. If one wanted this behavior, they could implement it locally in their project using this PR as a basis for what to do (and likely they would use FFI to get around the performance issue).

JordanMartinez:
I'm going to mark `arrays` as done in the core libs tracking issue

PR #209, opened 2021-01-18
Add changelog entries since v5.3.1

milesfrain:
**Description of the change**

This PR updates the CHANGELOG in preparation for the upcoming release of this library.

Related: purescript/purescript#3985

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Notable changes to this project are documented in this file. The format is based
 ## [Unreleased]
 
 Breaking changes:
+- Updated code for PureScript 0.14 (#181)
 - Renamed `Data.Array.ST.empty` to `Data.Array.ST.new` (#191 #198)
 - Renamed `group'` to `groupAll` (#194 #200)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,31 +5,30 @@ Notable changes to this project are documented in this file. The format is based
 ## [Unreleased]
 
 Breaking changes:
-- Updated code for PureScript 0.14 (#181)
-- Renamed `Data.Array.ST.empty` to `Data.Array.ST.new` (#191 #198)
-- Renamed `group'` to `groupAll` (#194 #200)
+- Added support for PureScript 0.14 and dropped support for all previous versions (#181)
+- Renamed `Data.Array.ST.empty` to `Data.Array.ST.new` (#191, #198)
+- Renamed `group'` to `groupAll` (#194, #200)
 
 New features:
-- Added monomorphic `foldl` `foldr` `foldMap` `fold` `intercalate` to `Array` (#201)
-  - These are just wrapped versions from `Data.Foldable`
-- Added monomorphic `foldl1` `foldr1` `foldMap1` `foldl1` `intercalate` to `Array.NonEmpty` (#201)
-  - These are just wrapped versions from `Data.Foldable`
-- Added monomorphic `elem` `notElem` `find` `findMap` `scanl` `scanr` `any` `all` (#189 #193 #201)
-- Added `intersperse` `groupAllBy` `splitAt`  (#179 #188 #194 #200 #201)
-- Added `mapWithIndex` `groupBy` to `Array.NonEmpty` (#201 #164)
+- Added specialized monomorphic functions from `Data.Foldable` (#201):
+  - Added `foldl`, `foldr`, `foldMap`, `fold`, `intercalate` to `Array` 
+  - Added `foldl1`, `foldr1`, `foldMap1`, `foldl1`, `intercalate` to `Array.NonEmpty`
+- Added monomorphic `elem`, `notElem`, `find`, `findMap`, `scanl`, `scanr`, `any`, `all` (#189, #193, #201)
+- Added `intersperse`, `groupAllBy`, `splitAt` (#179, #188, #194, #200, #201)
+- Added `mapWithIndex`, `groupBy` to `Array.NonEmpty` (#201, #164)
 
 Bugfixes:
-- Fixed `sort` so `undefined` is sorted by comparison function, and not simply moved to the end of the array (#195 #197)
+- Fixed `sort`, so `undefined` is sorted by comparison function and not simply moved to the end of the array (#195, #197)
 
 Other improvements:
 - Generated changelog and added PR template (#208, #209)
 - Added benchmarking (#178)
-- Migrated to GitHub Actions for CI (#187 #169)
+- Migrated to GitHub Actions for CI (#187, #169)
 - Removed some internal usages of `unsafeCoerce` (#184)
 - Changed `foldM` type signature to more closely match `foldl` (#160)
 - Updated installation instructions to use Spago (#171) 
 - Replaced foreign `cons` `snoc` `drop` `take` with purescript implementations (#180)
-- Removed `return {}` from FFI function (#175) 
+- Removed `return {}` from FFI function for a small performance boost (#175) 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,12 +5,33 @@ Notable changes to this project are documented in this file. The format is based
 ## [Unreleased]
 
 Breaking changes:
+- Updated code for PureScript 0.14 (#181)
```
thomashoneyman, line 8: > Added support for PureScript 0.14 and dropped support for all previous versions

CHANGELOG.md
```diff
@@ -5,12 +5,33 @@ Notable changes to this project are documented in this file. The format is based
...
+- Renamed `Data.Array.ST.empty` to `Data.Array.ST.new` (#191 #198)
+- Renamed `group'` to `groupAll` (#194 #200)
 
 New features:
+- Added monomorphic `foldl` `foldr` `foldMap` `fold` `intercalate` to `Array` (#201)
+  - These are just wrapped versions from `Data.Foldable`
```
thomashoneyman, line 14: I think these should be comma-separated:

```
- Added monomorphic `foldl`, `foldr`, ... and `intercalate` ...
```

CHANGELOG.md
```diff
@@ -5,12 +5,33 @@ Notable changes to this project are documented in this file. The format is based
...
+- Added monomorphic `foldl1` `foldr1` `foldMap1` `foldl1` `intercalate` to `Array.NonEmpty` (#201)
```
thomashoneyman, line 15: Should be comma-separated

CHANGELOG.md
```diff
@@ -5,12 +5,33 @@ Notable changes to this project are documented in this file. The format is based
...
+  - These are just wrapped versions from `Data.Foldable`
+- Added monomorphic `elem` `notElem` `find` `findMap` `scanl` `scanr` `any` `all` (#189 #193 #201)
```
thomashoneyman, line 17: Should be comma-separated -- the pull request numbers as well

CHANGELOG.md
```diff
@@ -5,12 +5,33 @@ Notable changes to this project are documented in this file. The format is based
...
 Breaking changes:
+- Updated code for PureScript 0.14 (#181)
+- Renamed `Data.Array.ST.empty` to `Data.Array.ST.new` (#191 #198)
+- Renamed `group'` to `groupAll` (#194 #200)
 
 New features:
+- Added monomorphic `foldl` `foldr` `foldMap` `fold` `intercalate` to `Array` (#201)
+  - These are just wrapped versions from `Data.Foldable`
+- Added monomorphic `foldl1` `foldr1` `foldMap1` `foldl1` `intercalate` to `Array.NonEmpty` (#201)
+  - These are just wrapped versions from `Data.Foldable`
```
thomashoneyman, line 16: I'm not sure "wrapped versions" is the correct terminology here; @hdgarrood, do you have a suggestion?

hdgarrood: Yeah, I think “wrapped” suggests that actually you’ve done something to them, whereas all we’ve done here is specialize. I’d suggest “specialized versions of the functions with the same names from Data.Foldable”. Is that true for all of these though? This shouldn’t be a separate bullet point either, it’s part of the same change.

hdgarrood: To be clear, the one I’m not sure about is intercalate - I thought it was supposed to get its own Array-specific implementation.

thomashoneyman: > This shouldn’t be a separate bullet point either, it’s part of the same change.

It's indented, to indicate a clarification rather than an entire new point, but I think it reads a little better as just sentences in the same bullet point and agree it should just merge into the previous line.

CHANGELOG.md
```diff
@@ -5,12 +5,32 @@ Notable changes to this project are documented in this file. The format is based
...
 
 Breaking changes:
+- Added support for PureScript 0.14 and dropped support for all previous versions (#181)
+- Renamed `Data.Array.ST.empty` to `Data.Array.ST.new` (#191, #198)
+- Renamed `group'` to `groupAll` (#194, #200)
 
 New features:
+- Added specialized monomorphic functions from `Data.Foldable` (#201):
+  - Added `foldl`, `foldr`, `foldMap`, `fold`, `intercalate` to `Array` 
+  - Added `foldl1`, `foldr1`, `foldMap1`, `foldl1`, `intercalate` to `Array.NonEmpty`
```
milesfrain, line 15: Reorganized the nested bullets to cut down on duplication. What's the follow-up step for `intercalate` mentioned in the earlier discussion?

CHANGELOG.md
```diff
@@ -5,12 +5,32 @@ Notable changes to this project are documented in this file. The format is based
 ## [Unreleased]
 
 Breaking changes:
+- Added support for PureScript 0.14 and dropped support for all previous versions (#181)
+- Renamed `Data.Array.ST.empty` to `Data.Array.ST.new` (#191, #198)
+- Renamed `group'` to `groupAll` (#194, #200)
 
 New features:
+- Added specialized monomorphic functions from `Data.Foldable` (#201):
```
hdgarrood, line 13: I actually find this a bit confusing, because these functions aren’t actually from Data.Foldable, they’re versions of those functions. I’d suggest “Added specialized versions of the functions from Data.Foldable”. Saying they’re specialized is the same thing as saying they’re monomorphic, and I think using both words makes this a bit harder to parse.

CHANGELOG.md
hdgarrood: The question is whether intercalate should get its own Array specific implementation (for performance) rather than just being a specialization of the intercalate from Data.Foldable. I think it probably should, or at least we should measure to see if an Array-based implementation would speed it up, but I suppose that doesn’t need to happen right now.

CHANGELOG.md
JordanMartinez: What about
> Exported monomorphic functions from `Data.Foldable` (i.e. `f` is forced to be `Array`)

It doesn't include the `HeytingAlgebra` being forced to be `Boolean` changes we also made (e.g. `all`, `any`), but that's the heart of it.

hdgarrood: I still prefer to say “Added specialized versions of the functions from Data.Foldable“. My problem is with saying the functions are from Data.Foldable (as opposed to being versions of the functions from Data.Foldable).

CHANGELOG.md
JordanMartinez: I've opened #210 to track this.

CHANGELOG.md
```diff
@@ -5,12 +5,32 @@ Notable changes to this project are documented in this file. The format is based
...
+- Fixed `sort`, so `undefined` is sorted by comparison function and not simply moved to the end of the array (#195, #197)
 
 Other improvements:
+- Generated changelog and added PR template (#208, #209)
+- Added benchmarking (#178)
+- Migrated to GitHub Actions for CI (#187, #169)
+- Removed some internal usages of `unsafeCoerce` (#184)
+- Changed `foldM` type signature to more closely match `foldl` (#160)
+- Updated installation instructions to use Spago (#171) 
+- Replaced foreign `cons`, `snoc`, `drop`, `take` with purescript implementations (#180)
```
JordanMartinez, line 30: Minor fix with capitalization:
```diff
- purescript
+ PureScript
```

JordanMartinez (changes requested):
I think only one minor change needs to be made before this can be merged.

Thanks for tackling this @milesfrain!

thomashoneyman (approved):

JordanMartinez (approved):

PR #212, opened 2021-04-13
added sliding-window functions sliding and slidingWithSizeAndStep

FloWi:
Added the functions sliding and slidingWithSizeAndStep to allow sliding-window computations. In order to realise the slidingWithSizeAndStep I added rangeWithStep.

I decided to have the functions return empty arrays in case the user gave them nonsensical arguments. Can change that to Maybe if that's preferred.

Thanks for the review, @sigma-andex.

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -33,6 +33,7 @@ module Data.Array
   , singleton
   , (..), range
   , rangeWithStep
+  , rangeWithStep'
   , replicate
   , some
   , many
@@ -89,7 +90,7 @@ module Data.Array
   , scanr
 
   , sliding
-  , slidingWithSizeAndStep
+  , slidingSizeStep
   , sort
   , sortBy
   , sortWith
@@ -1308,23 +1309,19 @@ sliding l = zip l (drop 1 l)
 -- |
 -- | ```purescript
 -- | > import Data.Array (range)
--- | > slidingWithSizeAndStep 3 2 (range 0 10) = [[0,1,2],[2,3,4],[4,5,6],[6,7,8],[8,9,10],[10]]
--- | > slidingWithSizeAndStep 3 3 (range 0 10) = [[0,1,2],[3,4,5],[6,7,8],[9,10]]
+-- | > slidingSizeStep 3 2 (range 0 10) = [[0,1,2],[2,3,4],[4,5,6],[6,7,8],[8,9,10],[10]]
+-- | > slidingSizeStep 3 3 (range 0 10) = [[0,1,2],[3,4,5],[6,7,8],[9,10]]
 -- | ```
 -- |
-slidingWithSizeAndStep :: forall a. Int -> Int -> Array a -> Array (Array a)
-slidingWithSizeAndStep size step array =
-  let
-    maxIndex = (length array) - 1
+slidingSizeStep :: forall a. Int -> Int -> Array a -> Array (NonEmptyArray a)
+slidingSizeStep size step array 
+  | size <= 0 || step <= 0 = [] {- args not valid -}
+  | otherwise = 
+    let
+      maxIndex = (length array) - 1
+    in
+      rangeWithStep' 0 maxIndex step (\i -> NonEmptyArray (slice i (i + size) array))
 
-    indices = rangeWithStep 0 maxIndex step
-
-    isValid = size > 0 && step > 0
-  in
-    if isValid then
-      indices <#> \i -> slice i (i + size) array
-    else
-      []
 
 -- | Create an array containing a range of integers with a given step size, including both endpoints.
 -- | Illegal arguments result in an empty Array.
@@ -1335,29 +1332,39 @@ slidingWithSizeAndStep size step array =
 -- | ```
 -- |
 rangeWithStep :: Int -> Int -> Int -> Array Int
-rangeWithStep start endIndex step =
-  let
-    isValid =
-      step /= 0
-        && if endIndex >= start then
-            step > 0
-          else
-            step < 0
+rangeWithStep start end step = rangeWithStep' start end step identity
 
-    hasReachedEnd curr =
-      if step > 0 then
-        curr > endIndex
-      else
-        curr < endIndex
+-- | Helper function to produce an array of elements like `rangeWithStep` with start, end and step, but immediatelly mapping over the result to work on one single mutable array.
+-- |
+-- | ```purescript
+-- | > rangeWithStep' 0 6 2 identity = [0,2,4,6]
+-- | > rangeWithStep' 0 6 2 (add 3) = [3,5,7,9]
+-- | ```
+rangeWithStep' :: forall t. Int -> Int -> Int -> (Int -> t) -> Array t
+rangeWithStep' start end step fn =
+  STA.run
+    ( do
+        let
+          isValid =
+            step /= 0
+              && if end >= start then
+                  step > 0
+                else
+                  step < 0
 
-    helper :: Int -> Array Int -> Array Int
-    helper currentIndex acc =
-      if hasReachedEnd currentIndex then
-        acc
-      else
-        helper (currentIndex + step) (snoc acc currentIndex)
-  in
-    if isValid then
-      helper start []
-    else
-      []
+          hasReachedEnd current =
+            if step > 0 then
+              current > end
+            else
+              current < end
+
+          helper current acc =
+            if hasReachedEnd current then
+              pure acc
+            else do
+              void $ STA.push (fn current) acc
+              helper (current + step) acc
+        arr <- STA.new
+        void $ helper start arr
+        pure arr
+    )
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -187,6 +187,7 @@ singleton :: forall a. a -> Array a
 singleton a = [a]
 
 -- | Create an array containing a range of integers, including both endpoints.
+-- | If you want to control the step-size, see `rangeWithStep`
 -- | ```purescript
 -- | range 2 5 = [2, 3, 4, 5]
 -- | ```
@@ -1294,8 +1295,9 @@ unsafeIndex = unsafeIndexImpl
 
 foreign import unsafeIndexImpl :: forall a. Array a -> Int -> a
 
--- | Takes an arrays of length n and returns an array of Tuples with length n-1.
--- | It's a "sliding window" view of this array with a window size of 2 and a step size of 1.
+-- | Returns an array where each value represents a "sliding window" of the given array with a window of size 2 and a step size of 1. 
+-- |
+-- | If you need more control over the size of the window and how far to step before making a new window, see `slidingSizeStep`
 -- |
 -- | ```purescript
 -- | sliding [1,2,3,4,5] = [(Tuple 1 2),(Tuple 2 3),(Tuple 3 4),(Tuple 4 5)]
@@ -1307,10 +1309,13 @@ sliding l = zip l (drop 1 l)
 -- | Takes an arrays and returns an array of arrays as a "sliding window" view of this array.
 -- | Illegal arguments result in an empty Array.
 -- |
+-- | As you can see in the example below, the last window might not get filled completely. 
+-- | Its size `s` will be `1 <= s <= size`.
+-- |
 -- | ```purescript
 -- | > import Data.Array (range)
--- | > slidingSizeStep 3 2 (range 0 10) = [[0,1,2],[2,3,4],[4,5,6],[6,7,8],[8,9,10],[10]]
--- | > slidingSizeStep 3 3 (range 0 10) = [[0,1,2],[3,4,5],[6,7,8],[9,10]]
+-- | > slidingSizeStep 3 2 [0,1,2,3,4,5,6,7,8,9,10] = [[0,1,2],[2,3,4],[4,5,6],[6,7,8],[8,9,10],[10]]
+-- | > slidingSizeStep 3 3 [0,1,2,3,4,5,6,7,8,9,10] = [[0,1,2],[3,4,5],[6,7,8],[9,10]]
 -- | ```
 -- |
 slidingSizeStep :: forall a. Int -> Int -> Array a -> Array (NonEmptyArray a)
@@ -1334,37 +1339,37 @@ slidingSizeStep size step array
 rangeWithStep :: Int -> Int -> Int -> Array Int
 rangeWithStep start end step = rangeWithStep' start end step identity
 
--- | Helper function to produce an array of elements like `rangeWithStep` with start, end and step, but immediatelly mapping over the result to work on one single mutable array.
+-- | Works just like rangeWithStep, but also uses a function to map each integer.
+-- | `rangeWithStep' start end step f` is the same as `map f $ rangeWithStep start end step`,
+-- | but without the extra intermediate array allocation.
 -- |
 -- | ```purescript
 -- | > rangeWithStep' 0 6 2 identity = [0,2,4,6]
 -- | > rangeWithStep' 0 6 2 (add 3) = [3,5,7,9]
+-- | > rangeWithStep' 0 (-6) (-2) (add 3) = [3,1,-1,-3]
 -- | ```
 rangeWithStep' :: forall t. Int -> Int -> Int -> (Int -> t) -> Array t
 rangeWithStep' start end step fn =
-  STA.run
-    ( do
-        let
-          isValid =
-            step /= 0
-              && if end >= start then
-                  step > 0
-                else
-                  step < 0
-
-          hasReachedEnd current =
-            if step > 0 then
-              current > end
-            else
-              current < end
-
-          helper current acc =
-            if hasReachedEnd current then
-              pure acc
-            else do
-              void $ STA.push (fn current) acc
-              helper (current + step) acc
-        arr <- STA.new
-        void $ helper start arr
-        pure arr
-    )
+  if not isValid 
+  then []
+  else STA.run do
+    let
+      helper current acc =
+        if hasReachedEnd current then
+          pure acc
+        else do
+          void $ STA.push (fn current) acc
+          helper (current + step) acc
+    arr <- STA.new
+    _ <- helper start arr
+    pure arr  
+  where 
+  isValid = 
+    step /= 0 && 
+      if end >= start 
+      then step > 0
+      else step < 0
+  hasReachedEnd current =
+    if step > 0 
+    then current > end
+    else current < end
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -32,6 +32,7 @@ module Data.Array
   , toUnfoldable
   , singleton
   , (..), range
+  , rangeWithStep
   , replicate
   , some
   , many
@@ -87,6 +88,8 @@ module Data.Array
   , scanl
   , scanr
 
+  , sliding
+  , slidingWithSizeAndStep
   , sort
   , sortBy
   , sortWith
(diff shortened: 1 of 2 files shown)


FloWi:
Hi maintainers! 
I'm quite new to the purescript-community (coming from scala) and was missing a function that comes in handy when analysing time series data. So I created a PR that adds these functions to arrays. Let me know what you think and if I can improve my code.

JordanMartinez:
Just a few initial thoughts about this:
- I don't see why we wouldn't add this. Adding a `sliding` view of an array is helpful. However, this sort of opens a can of worms because if we do so here, then we should likely add such API to both `lists` and their non-empty/lazy counterparts (though I'm not entirely sure about the lazy counterparts). That being said, I don't think we should wait until PRs adding that to other places are open before adding it here. 
- we could implement this without trying to be performant. For some use cases, this is "good enough", but we should document at least the Big O notation for these algorithms. If we want to be as fast as possible, then FFI would be preferred. However, that comes at the cost of making other backends implement this code since `STFn` as a concept can't be used in core libraries yet. If we used just `ST` to implement this, I'm not sure whether it would be more or less performant than the current implementation.
- On naming things, I think `slidingWithSizeAndStep` is a very long name and hope we can agree upon a shorter name that still communicates its idea. That might be as simple as renaming it to `sliding'`, a version of `sliding` with more configuration options shown. This name follows general conventions/patterns for naming things.
- `slidingWithSizeAndStep` should likely return `Array (NonEmptyArray a)` because when `step <= 0`, an empty array will be returned. Changing the return type to `NonEmptyArray` means `Foldable1` and its counterparts can be used in addition to `Foldable`.

FloWi:
Hi Jordan!

I started using purescript around one month ago and I love your great website with all the examples. Used it a lot while getting familiar with the language and it helped me a lot. 👍 

Thanks for your comprehensive input on the PR.

tl;dr Here's my plan - what do you think?

* introduce records to give the arguments names instead of currying? (if this is common for APIs in purescript). Don't like this signature-part `-> Int -> Int -> Int`
* create `rangeWithStep' :: forall a. (Int -> a) -> Int -> Int -> Int -> Array a` to use `ST.push` to avoid unnecessary copying
* implement `rangeWithStep = rangeWithStep' identity`
* rename `slidingWithSizeAndStep` to `sliding'`
* implement `sliding and sliding'` to use `rangeWithStep'` and pass the call to `slice` into it.
* change return type of `sliding'` to `Array (NonEmptyArray a)`

---

> * I don't see why we wouldn't add this. Adding a `sliding` view of an array is helpful. However, this sort of opens a can of worms because if we do so here, then we should likely add such API to both `lists` and their non-empty/lazy counterparts (though I'm not entirely sure about the lazy counterparts). That being said, I don't think we should wait until PRs adding that to other places are open before adding it here.

I think the runtime of a list-implementation would be quite horrible due to the lack of index-lookups - if the windows overlap. In that case it could be faster to copy the list into an array, perform the function and convert the result back to a list. So +1 for waiting until people ask for it. ;-)

> * we could implement this without trying to be performant. For some use cases, this is "good enough", but we should document at least the Big O notation for these algorithms. If we want to be as fast as possible, then FFI would be preferred. However, that comes at the cost of making other backends implement this code since `STFn` as a concept can't be used in core libraries yet.

I never heard of STFn. Do you mean this? <https://github.com/purescript/purescript-st> I also can't judge that part, since I'm new to purescript.

> If we used just `ST` to implement this, I'm not sure whether it would be more or less performant than the current implementation.

I use `snoc` for both creating my array of the starting-indices, and then again while mapping over the indices to collect the slices into the resulting array. Just saw that `snoc` always creates a mutable copy of the array. That is quite inefficient indeed. So, I like your idea of using ST to work on mutable arrays here. I think I'll rewrite it to use `St.push` on a mutable array. I can even rewrite my `rangeWithStep` function to also accept a `(Int -> a)` and push the result into the mutable array. That way I can reuse it with slice function and avoid the allocation of the intermediate array.

> * On naming things, I think `slidingWithSizeAndStep` is a very long name and hope we can agree upon a shorter name that still communicates its idea. That might be as simple as renaming it to `sliding'`, a version of `sliding` with more configuration options shown. This name follows general conventions/patterns for naming things.

Good to know, I like the idea. Is it ok to use a record to give the parameters names? Or is currying more common in purescript?

> * `slidingWithSizeAndStep` should likely return `Array (NonEmptyArray a)` because when `step <= 0`, an empty array will be returned. Changing the return type to `NonEmptyArray` means `Foldable1` and its counterparts can be used in addition to `Foldable`.

Great idea. Haven't thought about that.

hdgarrood:
I think we should avoid FFI in core because of portability - it should ideally not be any harder than it needs to be to get the core libraries running on a new backend.

I also prefer `slidingWithStepAndSize` over `sliding'`. I think primed names are fine for local variable names but not suitable for function names exported from a library, especially not a core library. All a prime tells you is that the function is slightly different from the non-primed version, but that’s not very helpful if you can’t remember how it’s different.

JordanMartinez:
> I started using purescript around one month ago and I love your great website with all the examples. Used it a lot while getting familiar with the language and it helped me a lot. +1

Glad to hear it helped!

> introduce records to give the arguments names instead of currying? (if this is common for APIs in purescript). Don't like this signature-part
> ...
> Good to know, I like the idea. Is it ok to use a record to give the parameters names? Or is currying more common in purescript?

I think what you have is good (i.e. currying). While it does make it harder to understand which argument is which, the alternatives aren't much better. Below are a few examples that highlight their problems:
1. One could pass in a record, but is it worth it to create a record object before consuming it pretty much immediately in the function just to fix this documentation issue? Probably not. Adding docs that indicate what the arguments are would likely be better.
2. One could define type alias that documents at the type signature what each type is (e.g. `type InclusiveStartIndex = Int`). However, type aliases sometimes do more harm than good because now you have the new problem of figuring out what `InclusiveStartIndex` is. For example, is that a zero-based index or a one-based index? Moreover, you need to export the type alias, but it doesn't really do anything. Perhaps this issue would be less important if a core library provided a number of common type aliases that others could use in their code for documentation purposes, but I'm not sure whether the tradeoffs are worth it. Lastly, at the end of the day, it's still just an `Int`, so one does not gain any type safety from it.

> I think the runtime of a list-implementation would be quite horrible due to the lack of index-lookups - if the windows overlap. In that case it could be faster to copy the list into an array, perform the function and convert the result back to a list. So +1 for waiting until people ask for it. ;-)

Yeah, terrible performance would likely cause us to go in that direction (i.e. transforming to an array first, sliding, and transforming back to a list).

> I never heard of `STFn`. Do you mean this? https://github.com/purescript/purescript-st I also can't judge that part, since I'm new to purescript.

`STFn` doesn't actually exist, but the idea does. There is `Effect`, that curries its arguments, and `EffectFn` that uncurries its arguments. For example, using arrow syntax for functions from JavaScript. 
- `Effect a` translates to `() => computationProducingA()`. 
- `a -> b -> c -> Effect d` translates to `(a) => (b) => (c) => () => computationProducingD(a, b, c)`, or a curried function, which isn't performance.
- To get an uncurried version of the above code, we'd use `EffectFn4 a b c d`. That translates to `(a, b, c) => () =>  computationProducingD(a, b, c)`.
- `ST h a` translates to `() => mutableComputationProducingA()`
- `a -> b -> c -> ST h d` translates to `(a) => (b) => (c) => () => mutableComputationProducingD(a, b, c)`
- We don't have an uncurried version of `ST`, which would be called `STFn4 h a b c d` in this context.

That's what the discussion around https://github.com/purescript/purescript-st/issues/31 involves (though there is more to it than just that). If that idea was implemented, we would get something as (nearly?) fast as FFI, but without the backend portability issues that Harry mentioned. However, this idea is more nuanced than what I'm summarizing here, so it's not just an "obvious" fix that we haven't yet implemented. 

Because `ST` is curried, it's not as fast as it could be if we got rid of the unneeded closures. I ran into a similar issue with performance when I tried implementing everything in `ST` in #203. I imagine yours would benefit if `ST` was used, but likely not as much as if an `STFn`-implementation could be used. I'd be curious to see the before/after benchmarks of your current implementation with an `ST`-based one.

> I think we should avoid FFI in core because of portability - it should ideally not be any harder than it needs to be to get the core libraries running on a new backend.

Also, because I don't think I was clear earlier, I agree with Harry on this point. Someone who really wants the same thing in a very performant way could write the FFI for their specific project.

> I also prefer `slidingWithStepAndSize` over `sliding'`. I think primed names are fine for local variable names but not suitable for function names exported from a library, especially not a core library.

I agree with this on principle, but I'd still like to see if we could make the name shorter. If we dropped with `with` and `and` part, would `slidingStepSize` be a good tradeoff? Or should we just stick with `slidingWithStepAndSize`?

src/Data/Array.purs
```diff
@@ -1289,3 +1292,72 @@ unsafeIndex :: forall a. Partial => Array a -> Int -> a
...
+    maxIndex = (length array) - 1
+
+    indices = rangeWithStep 0 maxIndex step
+
+    isValid = size > 0 && step > 0
+  in
+    if isValid then
+      indices <#> \i -> slice i (i + size) array
+    else
+      []
```
JordanMartinez, line 1327: Just a thought. Since PureScript is strict, `maxIndex` and, more importantly, `indices` will be calculated/produced even if the argument's aren't valid. You could prevent that from happening by doing this instead:
```
slidingWithSizeAndStep size step array
   | size <= 0 || step <= 0 = [] {- args not valid -}
   | otherwise = {- the rest of your let...in code -}
```

FloWi: Good to know. I really appreciate all your input!

FloWi:
Thanks for your input, @JordanMartinez and @hdgarrood. I really appreciate it. Always a pleasure when the first contact with the community of a programming language is positive.

I'm going to measure the current implementation and come up with improvements probably on the weekend and get back to you with results. Currently running a bit low on energy after work, since I have allergies and my archenemy (birch tree) is in full bloom.

JordanMartinez:
Sounds good!

FloWi:
It took me a while to figure out how to use `STA.new` and`STA.push`, since I wasn't able to write the signature of the array. Compiler was also complaining about scoping issues with the h parameter. I put my helper function now inside the `STA.run` and omitted the type-signature (type-inference ftw :)). Let me know if anything can be improved.

I skipped the measurement-part of the implementation, since it now operates on a mutable array and avoids copying and I didn't want to open another rabbit hole tbh ;-) Still at the stage in my learning curve where I struggle with the syntax from time to time.

src/Data/Array.purs
```diff
@@ -1289,3 +1293,78 @@ unsafeIndex :: forall a. Partial => Array a -> Int -> a
...
+-- |
+-- | ```purescript
+-- | sliding [1,2,3,4,5] = [(Tuple 1 2),(Tuple 2 3),(Tuple 3 4),(Tuple 4 5)]
+-- | ```
+-- |
+sliding :: forall a. Array a -> Array (Tuple a a)
+sliding l = zip l (drop 1 l)
+
+-- | Takes an arrays and returns an array of arrays as a "sliding window" view of this array.
+-- | Illegal arguments result in an empty Array.
```
JordanMartinez, line 1308: Perhaps add a comment in here that the last "window" size, `s` will be between `1 <= s <= size` when the given array does not have enough remaining elements to completely fill up the window?

src/Data/Array.purs
```diff
@@ -1289,3 +1293,78 @@ unsafeIndex :: forall a. Partial => Array a -> Int -> a
...
+-- |
+-- | ```purescript
+-- | > import Data.Array (range)
+-- | > slidingSizeStep 3 2 (range 0 10) = [[0,1,2],[2,3,4],[4,5,6],[6,7,8],[8,9,10],[10]]
+-- | > slidingSizeStep 3 3 (range 0 10) = [[0,1,2],[3,4,5],[6,7,8],[9,10]]
```
JordanMartinez, line 1313: Rather than using `range`, with which readers may not be familiar, thoughts on changing it to this?
```purescript
let array = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
slidingSizeStep 3 2 array = [[0,1,2],[2,3,4],[4,5,6],[6,7,8],[8,9,10],[10]]
```

src/Data/Array.purs
```diff
@@ -1289,3 +1293,78 @@ unsafeIndex :: forall a. Partial => Array a -> Int -> a
...
+-- |
+-- | ```purescript
+-- | > rangeWithStep 0 6 2 = [0,2,4,6]
+-- | > rangeWithStep 0 (-6) (-2) = [0,-2,-4,-6]
+-- | ```
+-- |
+rangeWithStep :: Int -> Int -> Int -> Array Int
+rangeWithStep start end step = rangeWithStep' start end step identity
+
+-- | Helper function to produce an array of elements like `rangeWithStep` with start, end and step, but immediatelly mapping over the result to work on one single mutable array.
```
JordanMartinez, line 1337: Could this sentence be split into multiple lines?

Also, I think this function's documentation could be clearer if it said something like
> Works just like `rangeWithStep`, but also uses a function to map each integer. 
> `rangeWithStep' start end step f` is the same as `map f $ rangeWithStep start end step array`
> but without the extra intermediate array allocation

src/Data/Array.purs
```diff
@@ -1289,3 +1293,78 @@ unsafeIndex :: forall a. Partial => Array a -> Int -> a
...
+rangeWithStep' start end step fn =
+  STA.run
+    ( do
+        let
+          isValid =
+            step /= 0
+              && if end >= start then
+                  step > 0
+                else
+                  step < 0
```
JordanMartinez, line 1353: Looks like `isValid` is no longer used in this function. Can this be removed?

test/Test/Data/Array.purs
```diff
@@ -467,6 +467,25 @@ testArray = do
     , [4,0,0,1,25,36,458,5842,23757]
     ]
 
+  log "sliding"
+  assert $ A.sliding [ 1, 2, 3, 4, 5 ] == [ (Tuple 1 2), (Tuple 2 3), (Tuple 3 4), (Tuple 4 5) ]
```
JordanMartinez, line 471: Can you add a test for `A.sliding [] == []`?

src/Data/Array.purs
```diff
@@ -1289,3 +1293,78 @@ unsafeIndex :: forall a. Partial => Array a -> Int -> a
...
+-- | Create an array containing a range of integers with a given step size, including both endpoints.
+-- | Illegal arguments result in an empty Array.
+-- |
+-- | ```purescript
+-- | > rangeWithStep 0 6 2 = [0,2,4,6]
+-- | > rangeWithStep 0 (-6) (-2) = [0,-2,-4,-6]
+-- | ```
+-- |
+rangeWithStep :: Int -> Int -> Int -> Array Int
+rangeWithStep start end step = rangeWithStep' start end step identity
```
JordanMartinez, line 1335: Could you add a sentence to the docs for `range` that says something like, "See also `rangeWithStep`"?

JordanMartinez (changes requested):
Overall looks good. Just need to clean up a few docs and some of the code.

src/Data/Array.purs
FloWi: My bad - I forgot to use it again during my refactoring. It's still needed to validate the arguments if you want to have a range "in the negative direction".

src/Data/Array.purs
```diff
@@ -1289,3 +1293,78 @@ unsafeIndex :: forall a. Partial => Array a -> Int -> a
 unsafeIndex = unsafeIndexImpl
 
 foreign import unsafeIndexImpl :: forall a. Array a -> Int -> a
+
+-- | Takes an arrays of length n and returns an array of Tuples with length n-1.
+-- | It's a "sliding window" view of this array with a window size of 2 and a step size of 1.
```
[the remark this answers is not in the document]
hdgarrood: I don't think we should explicitly address the case where someone modifies the array, because the `Array` type is immutable and the language doesn't give you any way of mutating them unless you explicitly opt in to unsafe mechanisms like ST's `unsafeFreeze/unsafeThaw` or the FFI.

src/Data/Array.purs
FloWi: done

test/Test/Data/Array.purs
FloWi: done

src/Data/Array.purs
FloWi: used it again and remembered, that purescript has the where block. Made it a bit more readable

src/Data/Array.purs
FloWi: done

src/Data/Array.purs
FloWi: done

src/Data/Array.purs
FloWi: good idea. The behaviour feels sometimes a bit weird (e.g. only one element), but in most cases it makes sense that it behaves that way. Let's hope, people don't get confused.

src/Data/Array.purs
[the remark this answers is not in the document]
FloWi: done

FloWi:
Thanks again for your input, @JordanMartinez and @hdgarrood. Made the documentation much clearer imo.

src/Data/Array.purs
```diff
@@ -1289,3 +1294,81 @@ unsafeIndex :: forall a. Partial => Array a -> Int -> a
...
+-- | ```purescript
+-- | sliding [1,2,3,4,5] = [(Tuple 1 2),(Tuple 2 3),(Tuple 3 4),(Tuple 4 5)]
+-- | ```
+-- |
+sliding :: forall a. Array a -> Array (Tuple a a)
+sliding l = zip l (drop 1 l)
+
+-- | Takes an arrays and returns an array of arrays as a "sliding window" view of this array.
+-- | Illegal arguments result in an empty Array.
+-- | As you can see in the example below, the last window might not get filled completely. Its size `s` will be `1 <= s <= size`.
```
JordanMartinez, line 1311: Can you break this into two lines?

src/Data/Array.purs
```diff
@@ -1289,3 +1294,81 @@ unsafeIndex :: forall a. Partial => Array a -> Int -> a
...
+-- | ```purescript
+-- | > rangeWithStep 0 6 2 = [0,2,4,6]
+-- | > rangeWithStep 0 (-6) (-2) = [0,-2,-4,-6]
+-- | ```
+-- |
+rangeWithStep :: Int -> Int -> Int -> Array Int
+rangeWithStep start end step = rangeWithStep' start end step identity
+
+-- | Works just like rangeWithStep, but also uses a function to map each integer.
+-- | `rangeWithStep' start end step f` is the same as `map f $ rangeWithStep start end step`, but without the extra intermediate array allocation
```
JordanMartinez, line 1341: Same here. Could you break it into two lines?

src/Data/Array.purs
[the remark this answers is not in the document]
JordanMartinez: > I don't think we should explicitly address the case where someone modifies the array

Yeah, that was likely an overreaction on my part because of the word, "view." The latest rendering looks good to me.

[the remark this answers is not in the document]
FloWi: No worries - aren't we all afraid of mutable stuff..? :P

src/Data/Array.purs
FloWi: Sure, all done.

src/Data/Array.purs
```diff
@@ -1334,37 +1339,39 @@ slidingSizeStep size step array
...
-              helper (current + step) acc
-        arr <- STA.new
-        void $ helper start arr
-        pure arr
-    )
+  if not isValid 
+  then []
+  else STA.run
+   ( do
+      let
```
JordanMartinez, line 1357: Just a stylistic change. The parenthesis after the `ST.run` can be removed if you follow it with `do`
```purescript
else ST.run do
  let
    helper current acc = {- code... -}
    
  arr <- STA.new
```
Just to clarify, normally `ST.run $ codeBlock` will work. However, because of the Rank-2 type in `ST.run` (i.e. the `h` type parameter), using `$` in this way will always produce a compiler error. However, using `do` creates a new block, so the parenthesis are no longer needed.

Other than that, I think this is ready to merge

JordanMartinez (approved):
I'll let other core team members weigh in on this PR.

One minor change that could also be done is to replace `void $` with `_ <-`. Not sure how much of a difference it makes. For example, in `rangeWithStep'`
```purescript
arr <- STA.new
-- Original
-- void $ helper start arr

-- Updated
_ <- helper start arr
pure arr

File: src/Data/Array.purs (hunk 11 above)
Comment (FloWi) [reply src/Data/Array.purs:1357]: I think I found that out the hard way when I tried to move the helper function out of the ST.run block. Never heard about rank-2 types before - at least not consciously ;-)

**Comment by FloWi** [comment]
The build failed because of an unused variable, but I haven't touched that place in the code. 🤔 
Were there some changes to the compiler-settings in CI? Couldn't find anything in the commits and the build runs fine on my machine.

**Comment by thomashoneyman** [comment]
[PureScript 0.14.1 was released yesterday](https://discourse.purescript.org/t/purescript-v0-14-1-released/2279) and we have CI on core libraries run on the latest compiler. The build is failing because of a newly-introduced warning. It's fixed by #213, so once that merges your build will be fixed as well.

**Comment by thomashoneyman** [comment]
You can merge the default branch to fix the build now. Thanks!

**Comment by FloWi** [comment]
Thanks, @thomashoneyman, will do later.

**Comment by hdgarrood** [comment]
Sorry that I didn't raise this sooner, but I am actually a bit hesitant to merge this at this moment in time. We ought to be really careful about adding things to the core libraries, because if we don't get the APIs for functions right the first time, it means we need to make breaking changes to fix them, and breaking changes in libraries that the overwhelming majority of the ecosystem depends on are much more painful than breaking changes that only a small section of the ecosystem depends on. In practice, I've found that it's almost impossible to get APIs right the first time, especially if we have not considered a variety of situations where the API might be useful to have.

Making sure the core libraries continue to improve and don't stagnate at the same time as avoiding unnecessary breakage and ecosystem churn is a very delicate balancing act. I'll suggest some rules of thumb for when something should be added to the core libraries based on my experience:

* There should be evidence of demand for API to be used relatively frequently, by different people in different situations,
* We should validate that the chosen design is appropriate in as many of these different situations as possible,
* We should have considered alternative designs for the API and be able to explain why the chosen design is more appropriate than those alternatives.

For an example of the last point - as I recall, the last time I wanted a sliding view over an array, I wanted something more like this:
```
type Window a = { previous :: Maybe a, current :: a, next :: Maybe a }
windows :: Array a -> Array (Window a)
```
where `length (windows xs) == length xs`.

Another reason to be wary about adding things to the core libraries is that it adds more of a maintenance burden on the core team. This is why I am often in favour of letting APIs be trialled in third party packages first - it is much easier to iterate on the design when the API lives in a separate package (because breaking changes can be made much more frequently), and it's also easier to provide evidence that the API is in demand if we see other packages start to depend on it. (as an aside: I think it might be nice to adopt a system like Rust's stability annotations which would make it much easier to iterate on things within the core libraries without introducing too much ecosystem churn, but that's probably quite a big project.)

I'm not saying we shouldn't merge this, I just think we should consider these points a bit more carefully before merging. I personally think I am spreading myself a bit too thin, and I'd like to focus more on the language and compiler, so in this specific case I am going to step back and let others in the core team decide (and I'll be happy with whatever your decision is).

JordanMartinez:
@hdgarrood Good thoughts! 

> as an aside: I think it might be nice to adopt a system like Rust's stability annotations which would make it much easier to iterate on things within the core libraries without introducing too much ecosystem churn, but that's probably quite a big project.

This sounds like a larger discussion to be had somewhere outside of this repo. I'll open an issue in the `purescript/governance` repo where we can discuss that more.

> For an example of the last point - as I recall, the last time I wanted a sliding view over an array, I wanted something more like this

Hmm... The Zipper-like design you propose seems similar to the approach I took when writing [purescript-arrays-zipper](https://github.com/JordanMartinez/purescript-arrays-zipper). But it seems like your approach is more flexible in that the resulting zipper could focus one element at a time or focus one window as done in this PR at a time. 

> I personally think I am spreading myself a bit too thin, and I'd like to focus more on the language and compiler, so in this specific case I am going to step back and let others in the core team decide (and I'll be happy with whatever your decision is).

Thanks! I think the direction I would like to take here is to continue with what you proposed, however, and think through this design more. In other words, I think it would make more sense to open a new issue on this repo, talk about the problem these functions try to solve, the various ways they could be designed, which approach makes the most sense, and whether that approach should exist as a separate library or be merged into here.

JordanMartinez:
@FloWi I'm going to open an issue where we can discuss the design of this idea more. I'll backlink to this PR in the issue.

JordanMartinez:
@FloWi See #214

sigma-andex:
Can this be merged?

sigma-andex:
@JordanMartinez can we merge this? I think it is better to have it than not and it implements a well established function that is also part of the standard libs in other programming languages, like e.g. Scala https://www.scala-lang.org/api/3.1.2/scala/collection/immutable/IndexedSeq.html#sliding-fffff156

JordanMartinez:
@sigma-andex Did you look at the discussion in #214? Could you continue the conversation there?

Issue #225, opened 2022-07-06
transpose?


newlandsvalley:
Is there any interest in adding ```transpose``` to arrays?  Not sure what would be a good implementation but possibly something along the lines of:

```purescript
transpose :: forall a. Array (Array a) -> Array (Array a)
transpose x = 
  case head x of 
    Just [_] -> 
      x
    _ -> 
      (mapMaybe head x) : transpose (mapMaybe tail x)
```

JordanMartinez:
Not sure if this is faster due to the allocations of arrays, but here's what I got:
```purs
transpose :: forall a. Array (Array a) -> Array (Array a)
transpose xs = go 0 []
 where
 go :: Int -> Array (Array a) -> Array (Array a)
 go idx allArrays = case buildNext idx of
  Nothing -> allArrays
  Just next -> go (idx + 1) (A.snoc allArrays next)
  
   
 buildNext :: Int -> Maybe (Array a)
 buildNext idx = do
  xs # flip F.foldl (Just []) \acc nextArr -> do
   el <- A.index nextArr idx
   nextArray <- acc
   pure $ A.snoc nextArray el
```
Which accounts for array elements that are unequal in size:
```
transpose [[1], [2], [3]] == [[1, 2, 3]]
transpose [[1], [2], [3, 4]] == [[1, 2, 3]]
transpose [[1, 2], [3, 4], [5, 6, 7]] == [[1, 3, 5], [2, 4, 6]]
```

newlandsvalley:
I've been making a start at looking at correctness.  One thing I'm not sure about is your last two examples.  By analogy to ```Data.List.transpose```:

```purescript 
( (1 : Nil) : (2 : Nil) : (3 : (4 : Nil)) : Nil) == ((1 : 2 : 3 : Nil) : (4 : Nil) : Nil)
( (1 : (2 : Nil)) :  (3 : (4 : Nil)) : (5:  (6 : (7 : Nil))) : Nil) == ((1 : 3 : 5 : Nil) : (2 : 4 : 6 : Nil) : (7 : Nil) : Nil)
```

JordanMartinez:
Huh... I guess the question here is which is the correct way. I support the List version is right because it was supported first.

newlandsvalley:
Yeah - I suppose we could see what Haskell does which might confirm things.

newlandsvalley:
[...]

PR #226, opened 2022-07-27
Add transpose to Data.Array

newlandsvalley:
**Description of the change**

It seems to be a good idea to add ```transpose``` to ```Data.Array``` given that we have it already in ```Data.List```.  This is discussed in https://github.com/purescript/purescript-arrays/issues/225.  The implementation was actually provided by @JordanMartinez for which many thanks.

I've slotted the code towards the end of the section marked ```transformations```.

---

**Checklist:**

- Added the change to the changelog's "Unreleased" section
- Linked to issue #225 
- Documentation style is basically the same as for ```Data.List.transpose```
- Tests basically identical to those for ```Data.List.transpose```

### DIFF
--- a/src/Data/Array.purs
+++ b/src/Data/Array.purs
@@ -786,13 +786,28 @@ intercalate = F.intercalate
 -- | For example,
 -- |
 -- | ```purescript
--- | transpose [[1,2,3], [4,5,6]] == [[1,4], [2,5], [3,6]]
+-- | transpose 
+-- |   [ [1, 2, 3]
+-- |   , [4, 5, 6]
+-- |   ] == 
+-- |   [ [1, 4]
+-- |   , [2, 5]
+-- |   , [3, 6]
+-- |   ]
 -- | ```
 -- |
 -- | If some of the rows are shorter than the following rows, their elements are skipped:
 -- |
 -- | ```purescript
--- | transpose [[10,11], [20], [30,31,32]] == [[10,20,30], [11,31], [32]]
+-- | transpose 
+-- |   [ [10, 11]
+-- |   , [20]
+-- |   , [30, 31, 32]
+-- |   ] == 
+-- |   [ [10, 20, 30]
+-- |   , [11, 31]
+-- |   , [32]
+-- |   ]
 -- | ```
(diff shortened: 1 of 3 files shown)


src/Data/Array.purs
```diff
@@ -781,6 +782,31 @@ fold = F.fold
...
+-- | For example,
+-- |
+-- | ```purescript
+-- | transpose [[1,2,3], [4,5,6]] == [[1,4], [2,5], [3,6]]
+-- | ```
+-- |
+-- | If some of the rows are shorter than the following rows, their elements are skipped:
+-- |
+-- | ```purescript
+-- | transpose [[10,11], [20], [30,31,32]] == [[10,20,30], [11,31], [32]]
```
JordanMartinez, line 795: Could examples docs be changed, so that the underlying matrix is better visualized? For example
```
transpose
  [ [ 1, 2, 3 ]
  , [ 4, 5, 6 ]
  ] ==
  [ [ 1, 4 ]
  , [ 2, 5 ]
  , [ 3, 6 ]
  ]

JordanMartinez: Or even
```purs
(transpose
  [ [ 1, 2, 3 ]
  , [ 4, 5, 6 ]
  ]) == (
  [ [ 1, 4 ]
  , [ 2, 5 ]
  , [ 3, 6 ]
  ])

newlandsvalley: OK - good idea.

JordanMartinez (approved):
I'm guessing `NonEmptyArray` also doesn't have `transpose`? Maybe that should be added as well. But, that can be done in a separate PR if you don't feel like doing it here.

JordanMartinez:
:ping_pong: @garyb Can this get an approval?

garyb (approved):
:+1:

newlandsvalley:
And ```NonEmptyList``` does not seem to have ```transpose``` either.  I suspect the main difficulty might be in writing terse documentation because constructing a ```NonEmptyArray``` is so awkward.  Maybe this is why this module is so lightly documented anyway.  Could we get away with a statement that says something like ```behaviour as for Array.transpose```?

Issue #227, opened 2022-07-30
Add transpose to NonEmptyArray


newlandsvalley:
I've had a look at modifying both implementations we have for the normal ```Array``` case detailed in #225 for use with a ```NonEmptyArray```.  I came up with this one based on my implementation:

```purescript
import Data.Array.NonEmpty
import Data.Array.NonEmpty.Internal (NonEmptyArray(..))
import Data.Maybe (Maybe(..))
import Prelude (map)

transpose :: forall a. NonEmptyArray (NonEmptyArray a) -> NonEmptyArray (NonEmptyArray a)
transpose x = 
  case (mapMaybe tail' x) of 
    [] -> 
      singleton (map head x)
    end -> 
      cons (map head x) (transpose (unsafeFromArray end))

  where
  tail' :: NonEmptyArray a -> Maybe (NonEmptyArray a)
  tail' x = 
    case tail x of 
      [] -> Nothing 
      end -> Just (unsafeFromArray end)

unsafeFromArray :: forall a. Array a -> NonEmptyArray a
unsafeFromArray = NonEmptyArray
```
and this one based on the implementation from @JordanMartinez :

```purescript
import Prelude
import Data.Array as A
import Data.Array.NonEmpty
import Data.Array.NonEmpty.Internal (NonEmptyArray(..))
import Data.Maybe (Maybe(..), maybe)
import Data.Foldable (foldl) 

transpose :: forall a. NonEmptyArray (NonEmptyArray a) -> NonEmptyArray (NonEmptyArray a)
transpose xs = unsafeFromArray $ go 0 []
  where
  go :: Int -> Array (NonEmptyArray a) -> Array (NonEmptyArray a)
  go idx allArrays = case buildNext idx of
    Nothing -> allArrays
    Just next -> go (idx + 1) (A.snoc allArrays next)
    
  buildNext :: Int -> Maybe (NonEmptyArray a)
  buildNext idx = 
    xs # flip foldl Nothing \acc nextArr -> do
      maybe acc (\el -> Just $ maybe (singleton el) (flip snoc el) acc) $ index nextArr idx

unsafeFromArray :: forall a. Array a -> NonEmptyArray a
unsafeFromArray = NonEmptyArray
```
[...]

PR #228, opened 2022-08-02
Add transpose and transpose' to NonEmptyArray

newlandsvalley:
**Description of the change**

Now that we have ```transpose``` in ```Array``` we can add two variants of it to ```NonEmptyArray```.  ```transpose``` handles the case where both the inner and outer arrays are non-empty.  ```transpose'``` handles the case where the inner arrays are typed as a standard ```Array```.  The ```transpose'``` implementation courtesy of @JordanMartinez .

For a discussion see #227.

---

**Checklist:**

- [ ] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")

### DIFF
--- a/src/Data/Array/NonEmpty.purs
+++ b/src/Data/Array/NonEmpty.purs
@@ -64,6 +64,8 @@ module Data.Array.NonEmpty
   , foldMap1
   , fold1
   , intercalate
+  , transpose
+  , transpose'
   , scanl
   , scanr
 
@@ -166,11 +168,21 @@ unsafeFromArray = NonEmptyArray
 unsafeFromArrayF :: forall f a. f (Array a) -> f (NonEmptyArray a)
 unsafeFromArrayF = unsafeCoerce
 
+-- Note that this is unsafe: if the array or any embedded array is empty, this can
+-- explode at runtime.
+unsafeFromArray2D :: forall a. Array (Array a) -> NonEmptyArray (NonEmptyArray a) 
+unsafeFromArray2D =  
+  (map unsafeFromArray) <<< unsafeFromArray
+
 fromNonEmpty :: forall a. NonEmpty Array a -> NonEmptyArray a
 fromNonEmpty (x :| xs) = cons' x xs
 
 toArray :: forall a. NonEmptyArray a -> Array a
 toArray (NonEmptyArray xs) = xs
+
+toArray2D :: forall a. NonEmptyArray (NonEmptyArray a) -> Array (Array a)
+toArray2D =
+  (map toArray) <<< toArray
 
 toNonEmpty :: forall a. NonEmptyArray a -> NonEmpty Array a
 toNonEmpty = uncons >>> \{head: x, tail: xs} -> x :| xs
@@ -360,6 +372,41 @@ fold1 = F.fold1
 
 intercalate :: forall a. Semigroup a => a -> NonEmptyArray a -> a
 intercalate = F.intercalate
+
+-- | The 'transpose' function transposes the rows and columns of its argument.
+-- | For example,
+-- |
+-- | ```purescript
+-- | transpose 
+-- |   (NonEmptyArray [ NonEmptyArray [1, 2, 3]
+-- |                  , NonEmptyArray [4, 5, 6]
+-- |                  ]) == 
+-- |   (NonEmptyArray [ NonEmptyArray [1, 4]
+-- |                  , NonEmptyArray [2, 5]
+-- |                  , NonEmptyArray [3, 6]
+-- |                  ])
+-- | ```
+-- |
+-- | If some of the rows are shorter than the following rows, their elements are skipped:
+-- |
+-- | ```purescript
+-- | transpose 
+-- |   (NonEmptyArray [ NonEmptyArray [10, 11]
+-- |                  , NonEmptyArray [20]
+-- |                  , NonEmptyArray [30, 31, 32]
+-- |                  ]) == 
+-- |   (NomEmptyArray [ NonEmptyArray [10, 20, 30]
+-- |                  , NonEmptyArray [11, 31]
+-- |                  , NonEmptyArray [32]
+-- |                  ])
+-- | ```
+transpose :: forall a. NonEmptyArray (NonEmptyArray a) -> NonEmptyArray (NonEmptyArray a)
+transpose = unsafeFromArray2D <<< A.transpose <<< toArray2D
+
+-- | `transpose`' is identical to `transpose` other than that the inner arrays are each
+-- | a standard `Array` and not a `NonEmptyArray`.
+transpose' :: forall a. NonEmptyArray (Array a) -> NonEmptyArray (Array a)
+transpose' = unsafeAdapt A.transpose
 
 scanl :: forall a b. (b -> a -> b) -> b -> NonEmptyArray a -> NonEmptyArray b
 scanl f x = unsafeAdapt $ A.scanl f x
--- a/src/Data/Array/NonEmpty.purs
+++ b/src/Data/Array/NonEmpty.purs
@@ -397,9 +397,10 @@ transpose =
     <<< A.transpose <<< coerce
 
 -- | `transpose`' is identical to `transpose` other than that the inner arrays are each
--- | a standard `Array` and not a `NonEmptyArray`.
-transpose' :: forall a. NonEmptyArray (Array a) -> NonEmptyArray (Array a)
-transpose' = unsafeAdapt A.transpose
+-- | a standard `Array` and not a `NonEmptyArray`. However, the result is wrapped in a 
+-- | `Maybe` to cater for the case where the inner `Array` is empty and must return `Nothing`.
+transpose' :: forall a. NonEmptyArray (Array a) -> Maybe (NonEmptyArray (Array a))
+transpose' = fromArray <<< A.transpose <<< coerce
 
 scanl :: forall a b. (b -> a -> b) -> b -> NonEmptyArray a -> NonEmptyArray b
 scanl f x = unsafeAdapt $ A.scanl f x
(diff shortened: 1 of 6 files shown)


src/Data/Array/NonEmpty.purs
```diff
@@ -166,12 +168,22 @@ unsafeFromArray = NonEmptyArray
...
+
 fromNonEmpty :: forall a. NonEmpty Array a -> NonEmptyArray a
 fromNonEmpty (x :| xs) = cons' x xs
 
 toArray :: forall a. NonEmptyArray a -> Array a
 toArray (NonEmptyArray xs) = xs
 
+toArray2D :: forall a. NonEmptyArray (NonEmptyArray a) -> Array (Array a)
+toArray2D =
+  (map toArray) <<< toArray
```
JordanMartinez, line 185: I believe you should be using `coerce` here (see https://pursuit.purescript.org/packages/purescript-safe-coerce/2.0.0/docs/Safe.Coerce#v:coerce). Otherwise, you're literally creating a copy of the original array with no change in the elements due to the usage of `map`. That's the very problem `coerce` solves.

That will require adding a dependency to `purescript-safe-coerce`.

JordanMartinez (changes requested):
My comment below also applies to `unsafeFromArray2D`.

src/Data/Array/NonEmpty.purs
garyb: That's a good note, but might beg the question of a larger change, since the current `unsafeAdapt`, and other related coercions would also probably benefit from `coerce`.

If it wasn't totally clear from Jordan's suggestion - the reason `coerce` is better than this suggested code is it can "deep coerce" without the need for mapping and such, hence no copying of array.

We could also just shoehorn an `unsafeCoerce` in, since that's already used in here. Would be nice to use `coerce` where we can though.

JordanMartinez: > We could also just shoehorn an `unsafeCoerce` in, since that's already used in here.

I disagree. `unsafeCoerce` made sense when we didn't have `Coercible` support. Now that we have support for it, let's design the code right.

newlandsvalley: So does this mean that both functions in question can just be implemented simple with ```coerce```   and that this is somehow safe so that ```unsafeFromArray2d``` could now be called ```fromArray2D``` ?

garyb: Yes and no :smile:

Both should be a single `coerce`, but there are two independent "safeties" involved here:

1. That he type being coerced is represented as an array. That's guaranteed by using `coerce`.
2. Whether the invariant that the array is actually non-empty is preserved, and we're not tracking that in the type system.

So going **to** `NonEmptyArray` is completely safe using `coerce`, going **from** `Array` we either need to check the invariants and make the function return `Maybe`, or mark it as unsafe, because it's up to the caller to have checked for non-emptiness first.

newlandsvalley: Yes I thought so - presumably ```coerce``` would have to rely on the safety of the ```NonEmptyArray``` constructor and this is manifestly unsafe.  I think I'm beginning to understand.  I hadn't realised before that coerce was deep, and so never thought that using it was a possibility.

newlandsvalley: I've made the change to use ```coerce```

src/Data/Array/NonEmpty.purs
```diff
@@ -361,6 +372,41 @@ fold1 = F.fold1
...
+-- |                  , NonEmptyArray [20]
+-- |                  , NonEmptyArray [30, 31, 32]
+-- |                  ]) == 
+-- |   (NomEmptyArray [ NonEmptyArray [10, 20, 30]
+-- |                  , NonEmptyArray [11, 31]
+-- |                  , NonEmptyArray [32]
+-- |                  ])
+-- | ```
+transpose :: forall a. NonEmptyArray (NonEmptyArray a) -> NonEmptyArray (NonEmptyArray a)
+transpose = unsafeFromArray2D <<< A.transpose <<< toArray2D
```
JordanMartinez, line 403: Since `toArray2D` is only used here (and same for `unsafeFromArray2D`), I would define these locally:
```purs
transpose = coerce <<< A.transpose <<< coerce
```

If the above doesn't compile, then I'd write:
```purs
transpose = (coerce :: _ (Array _) -> _ (NonEmptyArray _)) <<< A.transpose <<< (coerce :: _ (NonEmptyArray _) -> _ (Array _))
```
I don't think it's a good idea to expose the two other functions as it might encourage their use. I think such usages should be kept local so we can ensure the variant that the array is non-empty remains true.

newlandsvalley: Another good idea.  The ```coerce``` from ```fromArray2D``` required annotation, the other didn't.

src/Data/Array/NonEmpty.purs
```diff
@@ -361,6 +364,43 @@ fold1 = F.fold1
...
+transpose :: forall a. NonEmptyArray (NonEmptyArray a) -> NonEmptyArray (NonEmptyArray a)
+transpose = 
+  (coerce :: (Array (Array a)) -> (NonEmptyArray (NonEmptyArray a))) 
+    <<< A.transpose <<< coerce
+
+-- | `transpose`' is identical to `transpose` other than that the inner arrays are each
+-- | a standard `Array` and not a `NonEmptyArray`.
+transpose' :: forall a. NonEmptyArray (Array a) -> NonEmptyArray (Array a)
+transpose' = unsafeAdapt A.transpose
+
```
JordanMartinez, line 403: @garyb Just another thought, but should be there be a `transpose` variant when the input is `Array (NonEmptyArray a)`?

test/Test/Data/Array/NonEmpty.purs
```diff
@@ -397,6 +397,22 @@ testNonEmptyArray = do
...
+                         (fromArray [ (fromArray [10,20,30]), (fromArray [11,31]), (fromArray [32])])
+
+  log "transpose' swaps rows and columns for a regular two-dimension array"
+  assert $ NEA.transpose' (fromArray [[1,2,3], [4,5,6], [7,8,9]]) ==
+                          (fromArray [[1,4,7], [2,5,8], [3,6,9]])
+  
+  log "transpose' skips elements when rows don't match"
+  assert $ NEA.transpose' (fromArray [[10,11], [20], [30,31,32]]) ==
+                          (fromArray [[10,20,30], [11,31], [32]])
+
```
JordanMartinez, line 415: Can another test be added here for `NEA.transpose' (fromArray [[]]) == (fromArray [[]])`?

JordanMartinez (changes requested):
I think adding one more test should be done, but otherwise this looks good to go.

src/Data/Array/NonEmpty.purs
garyb: I did give that a moment's thought, but I figure we can skip it in favour of doing like `maybe [] (NEA.toArray <<< NEA.transpose) (NEA.fromArray xs)` if it arises.

test/Test/Data/Array/NonEmpty.purs
newlandsvalley: Hmm, this test fails - it produces ```(fromArray [])```.  Perhaps we need to special-case it?

src/Data/Array/NonEmpty.purs
JordanMartinez: Ah, good point!

test/Test/Data/Array/NonEmpty.purs
JordanMartinez: Oh, I think my example is just wrong. `Array` has the same behavior in its test: https://github.com/purescript/purescript-arrays/blob/master/test/Test/Data/Array.purs#L280

JordanMartinez: I guess the test should be `NEA.transpose' (fromArray [[]]) == (fromArray [])`

newlandsvalley: Surely ```fromArray []``` is an empty ```NonEmptyArray``` and thus eminently unsafe.

JordanMartinez: Oh gosh... :man_facepalming: Thanks!

Then the implementation should be
```purs
transpose' :: forall a. NonEmptyArray (Array a) -> Maybe (NonEmptyArray (Array a))
transpose' = fromArray <<< A.transpose <<< coerce
```

newlandsvalley: I'm not entirely happy with this.  From my perspective, ```transpose' (fromArray [[]])``` is a very odd thing to do anyway and I wouldn't have thought it would be entirely incorrect for it to be an idempotent operation.  But the cost of doing what you suggest  is to make all the normal cases more awkward by wrapping the result in the ```Maybe```.

JordanMartinez: Awkward or not, it is correct. And if that's the definition, then it's the same as:
```purs
fromArray <<< A.transpose <<< NEA.toArray
```
So, I would argue now that we shouldn't implement that here.

newlandsvalley: I'll happily do what you suggest if it is the correct definition.  Unless I hear otherwise from @garyb today I'll do it this way - I don't feel strongly about it in any case.

(However just to expand my approach a bit, I'd argue that the minimal result for ```transpose Array (Array a)``` is ```[]```, that for ```NonEmptyArray (NonEmptyArray a)``` is ```[[something]]``` and that for ```NonEmptyArray (Array a)``` is ```[[]]```.  There seems to be no equivalent function on ```NonEmptyList```` in the Haskell library. which might indicate correctness.)

garyb: The result of transposing `[[]]` being `[[]]` does sound pretty reasonable to me (in both non-or-empty cases), but I guess if we have evidence from elsewhere that that's not normal, then yeah having `transpose'` fail with `Nothing` is probably the best path forward.

JordanMartinez (approved):

garyb (approved):
I'll merge and release this later today, writing myself a script to make the process easier, and this will be the test case :wink:

garyb:
Released as v7.1.0

PR #231, opened 2022-08-22
Build an array directly, not indirectly via list

JordanMartinez:
**Description of the change**

I noticed that `fromFoldable`'s FFI will create an intermediate list that it will then unfold into an array. AFAICT, this is completely pointless because we can just build the initial array and then reverse it.

Does anyone know why this was implemented this way?

In my implementation, I changed the `foldr` to `foldl` because if the `Foldable f` is a `List`, `foldr` will first reverse the list before passing it to this function. So, we actually traverse the list 3 times:
1. reversing the initial list
2. folding over the arg passed to `fromFoldable` when constructing the intermediate list
3. folding over the intermediate list to build the array

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ New features:
 Bugfixes:
 
 Other improvements:
+- Speed up `fromFoldable` by removing intermediate list construction (#231 by @JordanMartinez)
 
 ## [v7.1.0](https://github.com/purescript/purescript-arrays/releases/tag/v7.1.0) - 2022-08-06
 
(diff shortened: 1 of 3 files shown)


natefaubion:
This was previously considered and rejected because there's no way to guarantee that the folding function is called only once on each entry, so it violates referential transparency.

natefaubion:
See https://github.com/purescript/purescript-arrays/pull/55#issuecomment-189734558

JordanMartinez:
Man.... that... bites.

PR #235, opened 2022-10-30
Update non-ST FFI to use Fn

JordanMartinez:
**Description of the change**

Updates non-`ST` FFI to use uncurried functions via `Fn` types

---

**Checklist:**


bower.json
```diff
@@ -35,4 +36,4 @@
     "purescript-const": "^6.0.0",
     "purescript-minibench": "^4.0.0"
   }
-}
+}
```
thomashoneyman, line 39:
```suggestion
}

```

thomashoneyman:
What is the motivation for this? Just to write more idiomatic JS in the FFI? Is this a change we're going to apply across core?

JordanMartinez:
Here's my rationale:
1. it produces stylistically cleaner JS output (no test needed)
1. Those doing `dev` work (i.e. just doing `purs compile` w/ no optimizations via `purs-backend-es`) _should_ get code that runs faster (test needed to confirm)
1. once these functions use `Fn*` functions, `purs-backend-es` with no other changes _should_ optimize more things (test needed to confirm)
1. once these functions use `Fn*` functions, `purs-backend-es` can be further updated to know how to evaluate some of these things during its optimization pass (pretty sure this is true).

thomashoneyman (approved):

natefaubion:
> once these functions use Fn* functions, purs-backend-es with no other changes should optimize more things (test needed to confirm)
once these functions use Fn* functions, purs-backend-es can be further updated to know how to evaluate some of these things during its optimization pass (pretty sure this is true).

Just as an FYI for changes like this, changing FFI calling conventions will potentially break optimizations. `purs-backend-optimizer` cases the calling convention against unsafeIndexImpl (curried vs uncurried matters), so currently the point release here very likely does not optimize at all with backend-es.

natefaubion:
That is, if it's important for the core team (I'm not saying it is) to preserve optimizations in tools like backend-optimizer, it's important that we actually do follow up and verify the impact of these changes.

PR #237, opened 2022-11-16
Avoid unnecessary array allocation

i-am-the-slime:
**Description of the change**

Avoids reusing `pushAll` to implement `push` which saves on array allocations.

Closes #236 

---

**Checklist:**

- [ ] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
- [x] Linked any existing issues or proposals that this pull request should close
- [ ] Updated or added relevant documentation
- [ ] Added a test for the contribution (if applicable)

PR #240, opened 2023-06-13
Add foreign implementation for push

### LINKED ISSUE #236: Implement `STArray.push` via proper FFI
Right now `push` does `pushAll [ x ]`.
One would use an `STArray` for good performance and this implementation undermines this.

i-am-the-slime:
**Avoid array allocation for push**

The previous implementation would create a new array every time
This one is a pretty faithful call into JS's `push`.

Supersedes #237 and closes #236 
---

**Checklist:**

- [x] Linked any existing issues or proposals that this pull request should close

### DIFF
--- a/src/Data/Array/ST.js
+++ b/src/Data/Array/ST.js
@@ -106,10 +106,6 @@ export const toAssocArrayImpl = function (xs) {
   return as;
 };
 
-export const push = function (a) {
-   return function (xs) {
-     return function () {
-       return xs.push(a);
-     };
-   };
- };
+export const pushImpl = function (a, xs) {
+  return xs.push(a);
+};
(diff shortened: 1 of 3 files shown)


src/Data/Array/ST.js
```diff
@@ -105,3 +105,11 @@ export const toAssocArrayImpl = function (xs) {
...
   return as;
 };
+
+export const push = function (a) {
+   return function (xs) {
+     return function () {
+       return xs.push(a);
+     };
+   };
+ };
```
thomashoneyman, line 115: Would you mind using the same `STFn` FFI format used elsewhere here? For example, this could be:

```suggestion
export const pushImpl = function (a, xs) {
  return xs.push(a);
};
```

which would be foreign-imported using `STFn2`, and then `push` would be implemented as:

```
push a = runSTFn2 pushImpl
```

i-am-the-slime: Oh sorry, I thought this wasn't done for ST yet.

